PR #121, opened 2014-10-16
Added support for cert-only login without user and password

p-kraszewski:
Added new DialTLS_CertAuth function for RabbitMQ configured for EXTERNAL auth with ssl_cert_login plugin for userless/passwordless logons.

As of now, amqp just silently hangs, when server presents authentication metod not supported by amqp (when pickSASLMechanism finds no suitable match). I guess it should bail out with some sort of "No common SASL methods found" message.

Hope this helps,
 best regards.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -30,8 +30,21 @@ func (me *PlainAuth) Response() string {
 	return fmt.Sprintf("\000%s\000%s", me.Username, me.Password)
 }
 
+// CertAuth for RabbitMQ-auth-mechanism-ssl.
+type CertAuth struct {
+}
+
+func (me *CertAuth) Mechanism() string {
+	return "EXTERNAL"
+}
+
+func (me *CertAuth) Response() string {
+	return fmt.Sprintf("\000*\000*")
+}
+
 // Finds the first mechanism preferred by the client that the server supports.
 func pickSASLMechanism(client []Authentication, serverMechanisms []string) (auth Authentication, ok bool) {
+
 	for _, auth = range client {
 		for _, mech := range serverMechanisms {
 			if auth.Mechanism() == mech {
--- a/connection.go
+++ b/connection.go
@@ -134,6 +134,25 @@ func DialTLS(url string, amqps *tls.Config) (*Connection, error) {
 		Heartbeat:       defaultHeartbeat,
 		TLSClientConfig: amqps,
 	})
+
+}
+
+// DialTLS_CertAuth accepts a string in the AMQP URI format and returns a new
+// Connection over TCP using EXTERNAL auth.  Defaults to a server heartbeat
+// interval of 10 seconds and sets the initial read deadline to 30 seconds.
+//
+// This mechanism is used, when RabbitMQ is configured for EXTERNAL auth with
+// ssl_cert_login plugin for userless/passwordless logons
+//
+// DialTLS_CertAuth uses the provided tls.Config when encountering an amqps://
+// scheme.
+func DialTLS_CertAuth(url string, amqps *tls.Config) (*Connection, error) {
+
+	return DialConfig(url, Config{
+		Heartbeat:       defaultHeartbeat,
+		TLSClientConfig: amqps,
+		SASL:            []Authentication{&CertAuth{}},
+	})
 }
 
 // DialConfig accepts a string in the AMQP URI format and a configuration for


michaelklishin:
In a discussion with @streadway and @gerhard we agreed to make this an example/expend the API to make existing `Dial*` functions to support more "advanced" parameters. The idea is to avoid `Dial*` function explosion.

michaelklishin:
Looks like `CertAuth` (or whatever we decide to name it) can be a part of this library because it won't change from user to user.

michaelklishin:
We agreed on `ExternalAuth` for struct name.

troian:
@streadway @michaelklishin wouldn't usage of Options pattern help to not fall with different suffixes to `Dial` func. (it's not exactly topic of current issue but touches it in a little).

what do I mean
```go
// Option callback for connection option
type Option func(*impl) error

// adding Options templates does not introduce compatibility issues
// as Dial(url) is still valid but adds flexibility in configuring AMQP connection as default
// configuration parameters still taking place
func Dial(url string, opts ...Option) (*Connection, error) {
	config := Config{
		Heartbeat: defaultHeartbeat,
		Locale:    defaultLocale,
	}

	return DialConfig(url, )
}

// SetOptions set amqp connection options
func (a *Config) SetOptions(opts ...Option) error {
	for _, opt := range opts {
		if err := opt(a); err != nil {
			return err
		}
	}

	return nil
}

func TLS(val *tls.Config) Option {
	return func(t * Config) error {
		t.TLSClientConfig = val
		return nil
	}
}

func Auth(val []amqp.Authentication) Option {
	return func(t *Config) error {
		t.SASL = val
		return nil
	}
}

// other parameters like connection timeout can be set through options too
```

from using prospective instead
```go
if t.cfg.TLS {
	conn = amqp.DialTLS(url, t.cfg.TLS)
} else {
	conn = amqp.Dial(url)
}
```

can be as
```go
conn = amqp.Dial(url, amqp.TLS(t.cfg.TLS), amqp.Auth(t.cfg.Auth))
```

All above is pseudo-code though we have done similar thing by making wrapper on top of `amqp.DialConfig`

prateek2408:
When is this getting merged

michaelklishin:
@prateek2408 @troian I don't use Go enough to tell if this pattern the way to go today. It is truly unfortunate that this PR has been sitting there for years. I would welcome any reasonable solution that would allow the users to use the `EXTERNAL` auth mechanism. @streadway do you have a preference?

auth.go
```diff
@@ -30,8 +30,21 @@ func (me *PlainAuth) Response() string {
...
+// CertAuth for RabbitMQ-auth-mechanism-ssl.
+type CertAuth struct {
+}
+
+func (me *CertAuth) Mechanism() string {
+	return "EXTERNAL"
+}
+
+func (me *CertAuth) Response() string {
+	return fmt.Sprintf("\000*\000*")
```
wolf-james, line 42: Change to `return "\000*\000*"`

auth.go
```diff
@@ -30,8 +30,21 @@ func (me *PlainAuth) Response() string {
 	return fmt.Sprintf("\000%s\000%s", me.Username, me.Password)
 }
 
+// CertAuth for RabbitMQ-auth-mechanism-ssl.
+type CertAuth struct {
```
wolf-james, line 34: Can we rename this `ExternalAuth` to match the RabbitMQ Java library naming convention?

auth.go
```diff
@@ -30,8 +30,21 @@ func (me *PlainAuth) Response() string {
...
+}
+
+func (me *CertAuth) Mechanism() string {
```
wolf-james, line 37: And for naming convention again, @streadway seems to use `auth` instead of `me`.

auth.go
```diff
@@ -30,8 +30,21 @@ func (me *PlainAuth) Response() string {
...
+	return "EXTERNAL"
+}
+
+func (me *CertAuth) Response() string {
+	return fmt.Sprintf("\000*\000*")
+}
+
 // Finds the first mechanism preferred by the client that the server supports.
 func pickSASLMechanism(client []Authentication, serverMechanisms []string) (auth Authentication, ok bool) {
+
```
wolf-james, line 47: Remove this unnecessary newline.

connection.go
```diff
@@ -134,6 +134,25 @@ func DialTLS(url string, amqps *tls.Config) (*Connection, error) {
...
+// Connection over TCP using EXTERNAL auth.  Defaults to a server heartbeat
+// interval of 10 seconds and sets the initial read deadline to 30 seconds.
+//
+// This mechanism is used, when RabbitMQ is configured for EXTERNAL auth with
+// ssl_cert_login plugin for userless/passwordless logons
+//
+// DialTLS_CertAuth uses the provided tls.Config when encountering an amqps://
+// scheme.
+func DialTLS_CertAuth(url string, amqps *tls.Config) (*Connection, error) {
+
```
wolf-james, line 150: Unnecessary newline

wolf-james:
@streadway Can we get this merged ASAP? I've been trying to get EXTERNAL to work for a few days now before finding this PR. @p-kraszewski's solution works perfectly for me, both with correct `user:pass` and `user:incorrect_pass`.

michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

PR #127, opened 2014-12-22
Fix #119

### LINKED ISSUE #119: Channel is not thread safe
Channel.shutdown(*error) function sets me.send function using lock. However Ack, Nack, Reject functions are not using the same lock.

Is the Channel supposed to be thread safe? Simple fix would be just add locking to those methods missing it, but I am not sure if there is something else behind this.

cenkalti:


cenkalti:
ping

streadway:
Thanks for the ping.  I'm not certain this fixes the race.  `shutdowned` is protected by `sendM` but that lock isn't held by `shutdown` where the state is written.

The lifecycle of channels changed in #128 which could move the open/closed handler to the connection rather than channel.  What do you think about an approach where "closing" is a concept that is coordinated outside of the channel?

cenkalti:
Sorry for late reply. My PR had passed the race detector test but your solution is also OK for me. Thanks for the fix.

streadway:
To be clear - the race may still exist (i haven't confirmed it yet).  I'm asking for your feedback on taking an approach that would move the channel.close/channel.close-ok handling up into the connection.  If so, then the channel would only have to worry about handling messages when it's considered open by the connection and channel closing can be synchronized on the connection.

cenkalti:
In my opinion opening/closing of channels should be handled by connection because a channel cannot open/close itself. It is better it should be handled by lower level (connection). I think your approach is fine. Sorry for the late reply.

PR #128, opened 2015-01-04
Negotiate and respect ChannelMax for Channel ID reuse

### LINKED ISSUE #115: Channel Error- Unsigned int rollover
I was running a server overnight that was reading in requests and sending them elsewhere. To do this I was using a pool of 200 connections. Each of these were making a channel for every request. So after 200*(2^16) straight requests, the channelRegistry number rolled over and tried to open channel 0 (a reserved RMQ channel) and crashed.

There should be a check when creating a new channel that .next() does not create channel 0

streadway:
This pull request adds an integer allocation and channel management strategy in order for channel IDs that wrap around uint16 not to conflict with previously allocated channel IDs or the connection ID (0).

This introduces a breaking API change in the `Config` struct.  I have tested the API breaking change against all packages listed in http://godoc.org/github.com/streadway/amqp?importers and for those that pass against master also pass against this branch.

Fixes #115 

@chb2mn @michaelklishin @carlhoerberg @w00204372 @chirimoya - please take a look


chirimoya:
I will have a look on wednesday ... thanks!

chirimoya:
+1 works fine for me

asim:
Ran into this issue the other day. Would like to see the fix get merged.

PR #130, opened 2015-01-20
On abnormal shutdown, avoid sending notification to already closed chans

echaz:
I was doing some testing of auto-reconnects with lots of data streaming through the amqp library.  This pull request adds a protection against a dead lock situation where some of the me.closes channels have already been closed.


echaz:
@streadway : ping?

streadway:
I'm interested in knowing how the channels become `nil` in the first place.  What behavior does this commit work around and do you have a test that can reproduce it?

PR #134, opened 2015-02-26
Properly handle basic.cancel from the server

joelegasse:
While working on a project that was consuming from queues that were getting deleted out-of-band from the client connection, I found out that RabbitMQ doesn't handle basic.cancel-ok sent from the client. Since this package was just blindly sending the cancel-ok and not closing the individual consumers, any consumer canceled by the server would result in the active connection getting closed... Not a pleasant way to go.

The two changes I made:

The client will no longer send a basic.cancel-ok message when no-wait is set in the basic.cancel message received from the server.

In addition, the consumer will be appropriately closed when the cancel is initiated by the server (due to queue deletion outside of the client)

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -275,7 +275,10 @@ func (me *Channel) dispatch(msg message) {
 		for _, c := range me.cancels {
 			c <- m.ConsumerTag
 		}
-		me.send(me, &basicCancelOk{ConsumerTag: m.ConsumerTag})
+		me.consumers.close(m.ConsumerTag)
+		if !m.NoWait {
+			me.send(me, &basicCancelOk{ConsumerTag: m.ConsumerTag})
+		}
 
 	case *basicReturn:
 		ret := newReturn(*m)


streadway:
Good find and the fix looks good.  Would you please include a regression test for each behavior - cleaning the consumer on cancel and honoring NoWait?

Thanks

zimmski:
I just stumbled upon this problem and it took me way too long to notice that there is already a fix. @streadway please give me some some guidance on what must be done to get his merged.

michaelklishin:
Thank you, I will include a test and correct one non-obvious thing: a `basic.cancel-ok` from the client is not required (Java and Ruby clients do not send them). This is not exactly coherent with how `basic.cancel` works in the opposite direction but let's do what the reference client implementation does.

zimmski:
@michaelklishin: Glad to see someone actively working on this package. Can you pinpoint me to the commit of yours which fixes the problem?

michaelklishin:
@zimmski I'm not done with tests yet, so nothing is pushed.

PR #141, opened 2015-04-27
Retry failed delivery on the next session properly

wy96f:
A "break" statement just terminates execution of the innermost "select", should break from innermost "for"


michaelklishin:
Thank you!

PR #143, opened 2015-05-23
Error out if exclusive and durable are used together

evanphx:
The docs on the function as well as the behavior within Rabbitmq indicate that exclusive implies non-durable and auto-delete.

Yet attempting to use exclusive and durable results in no error, just a non-durable queue available within RabbitMQ.

Better to error out if a user (like myself) attempts this than have them be puzzled until they notice that exclusive implies non-durable.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -7,6 +7,7 @@ package amqp
 
 import (
 	"container/heap"
+	"fmt"
 	"reflect"
 	"sync"
 )
@@ -752,6 +753,10 @@ declared with these parameters and the channel will be closed.
 func (me *Channel) QueueDeclare(name string, durable, autoDelete, exclusive, noWait bool, args Table) (Queue, error) {
 	if err := args.Validate(); err != nil {
 		return Queue{}, err
+	}
+
+	if exclusive && durable {
+		return Queue{}, fmt.Errorf("Invalid settings, can not be exclusive and durable")
 	}
 
 	req := &queueDeclare{


michaelklishin:
@evanphx your understanding is correct: exclusive queue lifecycle is attached to that of its declaring connection. FTR, it also means automatic deletion but under a different condition (connection closure rather than the last consumer going away).

evanphx:
@michaelklishin Do you think that the error is warranted then? Or is this behavior broker specific?

michaelklishin:
@evanphx it can be argued that this is a broker-specific behaviour but the spec entirely leaves out this detail. If a queue is deleted as soon as its declaring connection goes away/is closed, it cannot preserve broker restarts and it makes no sense to even try mirroring it.

None of the clients throw argument exceptions like the one in your PR but I wouldn't say it is unreasonable. Once you learn what various queue attributes mean, though, you are quite unlikely to be confused by this again.

So, I'd leave it up to the maintainer to decide :)

evanphx:
@michaelklishin ok, good to know. I haven't pushed AMQP enough to realize what 'exclusive' meant (which seems to be a meta-flag, implying auto-delete=true and durable=false). Sort of makes me wish the spec had not provided separate flags but instead a set of modes that ("exclusive", "durable", "durable-transient", etc).

michaelklishin:
@evanphx it means that the queue should be deleted as soon as its declaring connection is gone. Just another take on auto-deletion.

We will try clarifying this in the RabbitMQ docs.

michaelklishin:
@evanphx you probably have seen it but just in case: [AMQP 0-9-1 concepts](http://www.rabbitmq.com/tutorials/amqp-concepts.html).

evanphx:
@michaelklishin yup yup. I just didn't grok how much exclusive and durable + auto-delete are mutually exclusive flags to be used.

streadway:
I chose to defer most of these constraints to the server to decide whether or not a valid combination of flags makes sense. This library interface errs on the side of exposing the protocol rather than enforcing semantics. We have already seen some combinations around the immediate flag which became invalid with different RabbitMQ server versions so I'd rather leave that to the application's knowledge of where it is deployed.

streadway:
A documentation patch would be welcomed describing server behavior on mutually exclusive flag use.

PR #145, opened 2015-06-04
Change timestamp granularity to nanoseconds

telecoda:
** DO NOT MERGE ** Integration tests currently failing..

This changes the granularity of timestamps to be in nanoseconds rather seconds, as we were losing the accuracy in some header fields we were using.


telecoda:
Closing this, as it does not conform to the amqp protocol. Therefore will not interoperate with other amqp clients. (Will store my nanosecond timestamps as int64 values instead)

PR #158, opened 2015-09-01
Smooth out network/scheduling effects when sending heartbeats

evanphx:
The comment in the code explains it all. By sending the same rate that we emit heartbeats at, it's easy for network effects to cause the server to not see a heartbeat in time and time out the connection. Basically it turns into a distributed race condition.

By emitting the heartbeats at 75% of the rate the server expects, we have a much better chance of avoiding this situation.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -708,9 +708,20 @@ func (me *Connection) openTune(config Config, auth Authentication) error {
 		int(config.Heartbeat/time.Second),
 		int(tune.Heartbeat)))
 
+	// If we send our heartbeats at the same interval as the server
+	// is watching for them, any network or scheduling delays can
+	// easily cause the server to time the connection out.
+	//
+	// Thusly, to be safe, we tell the server the normal heartbeat rate
+	// but we actually send our heartbeats out at 75% of that rate, so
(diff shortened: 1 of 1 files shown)


evanphx:
This probably isn't actually necessary. RabbitMQ, at least, allows for 2 missed heartbeats before timing out the connection, so there is no need to skew the local emission time.

Feel free to close this if you don't feel it's necessary.

streadway:
Thanks for the concern! The higher rate would indeed be needed if RabbitMQ didn't handle multiple failed heartbeats.

This client tolerates 3 missed server heartbeats with the thought that having the same tolerable missed heartbeats as the server increases the chance of an idle disconnect race under symmetric congestion/partition. This client prefers the server to close first:  https://github.com/streadway/amqp/blob/master/connection.go#L517

I'm not certain how RabbitMQ behaves when applies TCP flow control on publishers. I assume that it also disables the heartbeat checks while TCP flow control (closed TCP window) is active. This client doesn't remove the read timeout under any TCP flow control on deliveries.

I can imagine head-of-the-line blocking from slow consumers that exceed 3 heartbeat intervals between deliveries will generate read timeouts which look like idle servers.

PR #162, opened 2015-09-09
special case fix for streadway/amqp/issues/161

trist4n:
fixes the situation in which both server and client have decided the
max frame size is 0 by detecting it and special casing the size to be
the length of what you're trying to publish, keeping the `Setting to 0
means "unlimited"` semantics from www.rabbitmq.com/configure.html


streadway:
Hi @trist4n - would you please add a test that exercises this code path? Thanks.

trist4n:
added a test, not entirely across how the server is being mocked, so i have used 'TestPublishBodySliceIssue74' as a template.

it seems to reproduce the crash on the master branch, and show the fix on my branch.

Issue #165, opened 2015-10-11
Missing TLS timeout


joliver:
During the TLS portion of the connection process, there is a missing timeout. Specifically, when the TLS [client is created and just before the handshake](https://github.com/streadway/amqp/blob/master/connection.go#L192), there needs to be a timeout set. Without this value there is an infinite timeout value which means the TLS handshake will stall.

The easy solution is to add a single statement:

https://golang.org/pkg/crypto/tls/#Conn.SetDeadline

`client.SetDeadline(config.TimeoutDuration) // or whatever configurable value you want`

Here's a stack trace to show you what's happening. We have observed this behavior when compiled against Go 1.4.2 and Go 1.5.2.  I have virtually identical stack traces from other machines.

```
goroutine 24 [IO wait, 125 minutes]:
net.(*pollDesc).Wait(0xc251acde90, 0x72, 0x0, 0x0)
    /usr/local/go/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitRead(0xc251acde90, 0x0, 0x0)
    /usr/local/go/src/net/fd_poll_runtime.go:89 +0x43
net.(*netFD).Read(0xc251acde30, 0xc24e6c3c00, 0x400, 0x400, 0x0, 0x7f6e3606ba48, 0xc23d3b6728)
    /usr/local/go/src/net/fd_unix.go:242 +0x40f
net.(*conn).Read(0xc23621c050, 0xc24e6c3c00, 0x400, 0x400, 0x0, 0x0, 0x0)
    /usr/local/go/src/net/net.go:121 +0xdc
crypto/tls.(*block).readFromUntil(0xc24a0abb60, 0x7f6e182f04b8, 0xc23621c050, 0x5, 0x0, 0x0)
    /usr/local/go/src/crypto/tls/conn.go:454 +0xe6
crypto/tls.(*Conn).readRecord(0xc24857c000, 0x16, 0x0, 0x0)
    /usr/local/go/src/crypto/tls/conn.go:539 +0x2da
crypto/tls.(*Conn).readHandshake(0xc24857c000, 0x0, 0x0, 0x0, 0x0)
    /usr/local/go/src/crypto/tls/conn.go:779 +0xe7
crypto/tls.(*Conn).clientHandshake(0xc24857c000, 0x0, 0x0)
    /usr/local/go/src/crypto/tls/handshake_client.go:141 +0x110a
[...]

PR #166, opened 2015-10-12
Ensuring the TLS handshake doesn't stall forever by providing a deadline

joliver:
This would resolve issue #165

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -190,6 +190,7 @@ func DialConfig(url string, config Config) (*Connection, error) {
 		}
 
 		client := tls.Client(conn, config.TLSClientConfig)
+		client.SetDeadline(defaultConnectionTimeout)
 		if err := client.Handshake(); err != nil {
 			conn.Close()
 			return nil, err


joliver:
@streadway Is there anything I can do to get this merged in? We've hit this bug a number of times in production now.

streadway:
Thanks for the reminder.

The `defaultDialer` already has [SetReadDeadline](https://github.com/smartystreets/amqp/blob/master/connection.go#L114)

Are you using a custom dialer? If not, please change the `defaultDialer` to `SetDeadline` since it looks like the TLS client handshake writes first.

joliver:
I'm using the `defaultDialer`.

Here's the trick. There are *two* deadlines that ned to be set&mdash;the dial timeout, which you have, and the TLS handshake timeout, which you're missing.

Ironically, I identified [this exact same issue](https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/5b03d351fe83c294021d3b18f71d369e78cb2017) in the official RabbitMQ .NET client library about two years ago. You can see my name in the comment for that commit.

streadway:
Yup - the TCP handshake deadline is handled in https://github.com/smartystreets/amqp/blob/master/connection.go#L108-L111

From there it's an optional TLS handshake and then the first messages of an AMQP handshake before heartbeating will check the liveliness of the connection.

It looks like SetReadTimeout in defaultDialer will timeout on the server sent TLS ServerHello read already. But if using TLS, it won't timeout on the AMQP ConnectionStart read.

I understand that Set*Timeout will wait for the next (single) IO operation, which means both the read and write timeout would be met on the first messages of the [TLS handshake](https://golang.org/src/crypto/tls/handshake_client.go#L140) before the AMQP handshake begins.  Since we don't control the timeouts between reads and writes within the tls package, I'm inclined to look at handshaking of TCP, TLS and AMQP under the same connection timeout.

This can be validated if TLS termination is separate from AMQP termination. TLS succeeds but AMQP stalls - the connection would hang.

It sounds like a better alternative would be to start the heartbeater as soon as the TCP session is established and reconfigure it with whatever the client and server negotiate during the AMQP connection handshake.

Do you think you could make the tests necessary to timeout on stalled TLS and also timeout on successful TLS but stalled AMQP to reproduce this bug?

joliver:
I definitely like the idea of a single timeout for the TCP, TLS, and AMQP handshake process. If we don't have a full AMQP connection, we should definitely tear down what we've got.

I might be able to create a reproducible test case, but I'll have to investigate how possible it is to mock the various aspects of the connection.

streadway:
Take a look at https://github.com/smartystreets/amqp/blob/master/tls_test.go which gets through a TLS handshake into an incomplete AMQP handshake.

From there you can look at https://golang.org/pkg/testing/iotest/#TruncateWriter to simulate a network partition.

michaelklishin:
We agreed to start small and use the same timeout value for each "stage": TCP connection, TLS upgrade, AMQP 0-9-1 handshake. Context propagation may come in the future.

michaelklishin:
@gerhard and I tried to simulate a TLS handshake timeout and after a good hour or so we have some findings:

 * Go's TLS implementation is good enough that it complains about specific issues such as oversized records (when a TLS client connects to a non-TLS server) or unknown CA certificate (non-TLS client to a TLS server, a common scenario that times out with some clients and I believe `openssl s_client`)
 * As far as I can tell, Go's standard library exposes only one timeout for each connection type, and while it's probably possible to configure TLS handshake timeout, it is not trivial to test just that part without modifying this client's code
 * We investigated how Go's HTTP transport implements TLS handshake timeouts for HTTPS. It does so in the transport's own code (by setting up a one-off timer), presumably because of the point above.

Therefore we believe further attempts at testing specifically the TLS handshake timeout is not worth the time it is taking. The patch is quite trivial and we've done the same thing in at least one official client in the past.

michaelklishin:
Manually applied in master, thank you @joliver!

PR #168, opened 2015-10-21
Travis: migrate to container-based infrastructure

wingyplus:


wingyplus:
Oh... I seem doesn't work with rabbitmq service.

PR #178, opened 2015-12-18
Added go tip to build matrix

njohns-grovo:
Hello,
Go v1.6 beta 1 was released today, I've added `tip` to the build matrix so all future go releases will be automatically tested without updating the `.travis.yml` file


michaelklishin:
Thank you!

PR #185, opened 2016-02-06
Expose tls.ConnectionState for TLS connections

cbandy:
I wanted to check the TLS version and cipher suite negotiated by Go and RabbitMQ.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -235,6 +235,16 @@ func (me *Connection) LocalAddr() net.Addr {
 	return &net.TCPAddr{}
 }
 
+// TLSConnectionState returns basic TLS details of the underlying transport.
+func (me *Connection) TLSConnectionState() tls.ConnectionState {
+	if c, ok := me.conn.(interface {
+		ConnectionState() tls.ConnectionState
+	}); ok {
+		return c.ConnectionState()
+	}
+	return tls.ConnectionState{}
+}
+
 /*
 NotifyClose registers a listener for close events either initiated by an error
 accompaning a connection.close method or by a normal shutdown.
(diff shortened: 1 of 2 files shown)


tls_test.go
```diff
@@ -93,6 +93,11 @@ func TestTLSHandshake(t *testing.T) {
 			t.Fatalf("expected to handshake a TLS connection, got err: %v", err)
 		}
 	}
+
+	st := c.ConnectionState()
+	if !st.HandshakeComplete {
+		t.Errorf("expected ConnectionState, got: %+v", st)
```
michaelklishin: This error message doesn't actually warn the user that there was a TLS handshake failure. Please change this to "TLS handshake failed, TLS connection state: …".

connection.go
```diff
@@ -232,6 +236,14 @@ func (me *Connection) LocalAddr() net.Addr {
 	return &net.TCPAddr{}
 }
 
+// ConnectionState returns basic TLS details of the underlying transport.
+func (me *Connection) ConnectionState() tls.ConnectionState {
```
michaelklishin: This function can confuse the reader slightly: it returns a `tls.ConnectionState` but AMQP 0-9-1 connections also have state and so `*Connection.ConnectionState` doesn't really indicate that what's returned in a TLS upgrade state.

Please rename it to `TLSConnectionState`.

michaelklishin:
Thank you! We'd be happy to include this feature but this specific PR needs a bit of work.

tls_test.go
cbandy: Fixed in amended commit.

connection.go
cbandy: Done.

Issue #150, opened 2015-07-20
channel.Confirm waits forever


wy96f:
Shall we return the err if waiting for the response times out?

Goroutine dump as follows:

goroutine 14 [select, 9635 minutes]:
github.com/streadway/amqp.(_Channel).call(0xc2081beb00, 0x7f326a25e2b8, 0xc20818fb28, 0xc20817bb10, 0x1, 0x1, 0x0, 0x0)
        /opt/go/go/src/github.com/streadway/amqp/channel.go:149 +0x5f1
github.com/streadway/amqp.(_Channel).Confirm(0xc2081beb00, 0x0, 0x0, 0x0)
        /opt/go/go/src/github.com/streadway/amqp/channel.go:1476 +0x1ba
match_server/mq.(*PubMessageQueue).publish(0xc208029840, 0xc20803e540, 0xc20803e5a0, 0xc208034410, 0x43)
        /opt/go/go/src/match_server/mq/pubsub.go:136 +0x171

michaelklishin:
Looks like there was a PR that addressed this.

PR #187, opened 2016-02-16
Avoid deadlock in Confirm()

abrander:
We should not hold the `me.m` lock when calling `call()`. It could result in a deadlock if the underlying connection calls `shutdown()` before the server replies. This PR fixes that.


abrander:
This could be related to the issue described in #150.

Issue #173, opened 2015-11-25
deadlock creating channel on closing connection


karlseguin:
We're seeing cases where opening a channel (`conn.Channel()`) hangs if the connection is in the process of shutting down. Specifically, the `select` in `channel.call` blocks.

It's hard to reproduce reliably, but if you make the following changes to `connection.go` (which helps extend the window in which this can happen), it becomes easy to reproduce.

1 - Make the following 2 changes to connections.go: https://github.com/2nd/amqp/commit/30a656f7430ce5ee6fa85c1efb52633741fe015f

2 - Run the following code

``` go
func main() {
    url := "amqp://guest:guest@localhost:5672"
    conn, err := amqp.Dial(url)
    if err != nil {
        log.Fatalf("connection.open: %s", err)
    }   

    wg := new(sync.WaitGroup)

    go func() {
        wg.Add(1)
        time.Sleep(time.Second)
        conn.Channel()
        wg.Done()
    }() 

    time.Sleep(time.Second)
    conn.Close()
    wg.Wait()
}
```

Our solution is to add a `closed` variable to the connection and reference it (always under lock) before allocating a channel. The code is https://github.com/2nd/amqp/commit/fb0a3d94dd9dde2bb44be49057d0adb512b4640c  (see commit message for more details).

We're not submitting this as a PR because we're just pushing it live today and don't feel confident pushing this on anyone else at this point. However, we'd welcome feedback on the issue & solution.

karlseguin:
Follow up:
While the commit solved the above code, we saw the error again. A follow up commit is currently looking more promising: https://github.com/2nd/amqp/commit/0b08b93c2157c997656d1e44c0519f03eb4121a8  (however, individually, this change didn't solve the issue either, so either the issue is still present, or only both changes together resolve it).

Channel's `call` method appears to handle `rpc` being closed, but nowhere does it appear to actually get closed. Looking through past commits, closing `rpc` as a means of breaking that select appears to have been replaced with writing to `errors` (though much of that code was kept around). If the issue is in fact resolved, it would appear to be indicate that the channel's shutdown is being closed without receiving an error, which doesn't quite seem possible to me...but, if it works...

Tangentially, if all else fails, we'll probably timeout that select. I rather deal with eager exits than no exits.

karlseguin:
Grokked.

The reason protecting the channels map, as seen in our first commit, didn't work is because of the following sequence:

Thread 1 - conn.Channnel()
Thread 1 - openChannel()
[...]

Issue #180, opened 2015-12-30
Open channel waits forever


joliver:
If the network connection to RabbitMQ is unstable (which can be the case over the Internet), we have found multiple times that our code gets stuck in the following location (see stack trace). We are using the default dialer which has a 30-second timeout along with the default 30-second read timeout. I'm investigating how to fix it, but I'm listing this here to document the problem.

Here's the stack trace:

```
goroutine 10 [select, 488 minutes]:
github.com/streadway/amqp.(*Channel).call(0xc21cd2eb00, 0x7f271dc471c0, 0xc24a168610, 0xc21a247c48, 0x1, 0x1, 0x0, 0x0)
    /users/buildserver/src/github.com/streadway/amqp/channel.go:149 +0x5f1
github.com/streadway/amqp.(*Channel).open(0xc21cd2eb00, 0x0, 0x0)
    /users/buildserver/src/github.com/streadway/amqp/channel.go:138 +0x170
github.com/streadway/amqp.(*Connection).openChannel(0xc21aafbe60, 0x0, 0x0, 0x0)
    /users/buildserver/src/github.com/streadway/amqp/connection.go:570 +0x7f
github.com/streadway/amqp.(*Connection).Channel(0xc21aafbe60, 0x1, 0x0, 0x0)
    /users/buildserver/src/github.com/streadway/amqp/connection.go:591 +0x3a
```

Here's the function in the `channel.go` file where the problem is manifest. It appears that the errors channel is never populated thus the select statement blocks forever.

```
func (me *Channel) call(req message, res ...message) error {
    if err := me.send(me, req); err != nil {
        return err
    }

    if req.wait() {
        select { // <-- WAITS RIGHT HERE FOREVER
        case e := <-me.errors:
            ...
        case msg := <-me.rpc:
            ...
        }
    }

    return nil
}
```

domashevskiy:
I think this might be the same as in [issue 173](https://github.com/streadway/amqp/issues/173).
There is a possible [fix](https://github.com/WaimaiChaoren/amqp/commit/88561e180aeee809b7cb48a164a71ebc2148c6fd) in a [fork](https://github.com/WaimaiChaoren/amqp) you can try.

karlseguin:
Just another update:

closing the rpc channel didn't seem to fix all the problems. Also, it's resulted in some code now trying to write to the closed channels. However, we've done other "fixes" to try to solve various other concurrency issues, so maybe it's related to one of those.

[...]

Issue #147, opened 2015-07-02
Deadlock when closing channels


nvx:
Out of habit, I am using "defer chan.Close()" after obtaining a channel.

Channels are returned from a separate goroutine that acts as a channel factory (handling reconnection all in one place). When I send a signal to the context to shutdown everything gracefully, I find my workers hung in chan.Close(). My suspicion is that conn.Close() is being called first (or simultaneously) and either conn.Close() followed by chan.Close() hangs, or conn.Close() running simultaneously with chan.Close() causes the latter to hang.

As a quick fix I just removed the chan.Close() and everything seems to be fine as conn.Close() appears to clean up channels anyway, but I feel like this should be fixed.

domashevskiy:
Sample code. It hangs in most cases. If you build it with "-race" flag you'll see data race details.
```go
package main

import (
	"fmt"
	"github.com/streadway/amqp"
	"log"
	"runtime"
	"sync"
	"time"
)

func init() {
	runtime.GOMAXPROCS(runtime.NumCPU())
}

func createChannels(conn *amqp.Connection, wg *sync.WaitGroup) {
	ch, err := conn.Channel()
	if err != nil {
		panic(err)
	}

	time.Sleep(time.Second)
	fmt.Println("Chan close")

	err = ch.Close()

	wg.Done()
}

func channels() {
	conn, err := amqp.Dial("amqp://login:pass@localhost:5672/")
	if err != nil {
		log.Fatalf("connection.open: %s", err)
	}

	wg := new(sync.WaitGroup)

	for i := 0; i < 8; i++ {
		wg.Add(1)
		go createChannels(conn, wg)
	}
	time.Sleep(time.Second)
	fmt.Println("Conn close")
	conn.Close()

	wg.Wait()
}

func main() {
	channels()
}
```

domashevskiy:
It hangs in channel.go's call function. It waits infinitely to receive something from me.rpc. Maybe you should close rpc channel on shutdown? 
In my experiments adding close(me.rpc) in shutdown method of a channel helped.
```go
// Performs a request/response call for when the message is not NoWait and is
// specified as Synchronous.
func (me *Channel) call(req message, res ...message) error {
	if err := me.send(me, req); err != nil {
		return err
	}

	if req.wait() {
		select {
		case e := <-me.errors:
			return e

		case msg := <-me.rpc:
			if msg != nil {
				for _, try := range res {
					if reflect.TypeOf(msg) == reflect.TypeOf(try) {
[...]

PR #188, opened 2016-02-16
Never send any frames after connection shutdown

abrander:
We should never send any frames after shutdown. Doing so can easily result in deadlocks.

This PR fixes this deadlock by using a `closed` field on the Connection type which will be set upon `shutdown()`.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -93,6 +94,8 @@ type Connection struct {
 	Major      int   // Server's major version
 	Minor      int   // Server's minor version
 	Properties Table // Server properties
+
+	closed int32 // Will be 1 if the connection is closed, 0 otherwise. Should only be accessed as atomic
 }
 
 type readDeadliner interface {
@@ -315,6 +318,10 @@ func (me *Connection) closeWith(err *Error) error {
 }
 
 func (me *Connection) send(f frame) error {
+	if atomic.LoadInt32(&me.closed) == 1 {
+		return ErrClosed
+	}
+
 	me.sendM.Lock()
 	err := me.writer.WriteFrame(f)
 	me.sendM.Unlock()
@@ -339,6 +346,8 @@ func (me *Connection) send(f frame) error {
 }
 
 func (me *Connection) shutdown(err *Error) {
+	atomic.StoreInt32(&me.closed, 1)
+
 	me.destructor.Do(func() {
 		if err != nil {
 			for _, c := range me.closes {


abrander:
I think this fixes issue #173 and #180. Maybe even #147 and #150.

lud4ik:
Interesting. Can you describe, please, scenarious that your commit fixes.
Did it fixed your case and have been tested?

abrander:
It can be reproduced fairly simple:

```go
connection := amqp.Dial(...)
for {
    chan, _ := connection.Channel()
    fmt.Printf("We're alive %p\n", chan)
    chan.Close()
}
```

Run a loop like this in a few hundred goroutines while restarting the peer (we did `rabbitctl stop_app && rabbitctl start_app` once a second). It would deadlock for us within a few seconds.

The race took some hard work and a lot of packet captures to find ;)

A function (In this case `Channel.open()` called from `Connection.Channel()`) calling `Channel.call()` when doing RPC will send a `*channelOpen` which will trigger `Channel.call()`'s `select{}` waiting for a reply.

Imagine this:
The serverside (or anyone else) closes the connection.
The client receives a `*connectionClose` and starts `Connection.shutdown()` immediately (in another goroutine).
Someone calls `Connection.Channel()` *before* `Connection.shutdown()` reaches `me.conn.Close()`.
The `*channelOpen` frame *will* be sent and `Channel.rpc()` will be waiting for a reply or error that will never arrive.
Boom. Deadlock in `Channel.call()`.

All RPC functions are susceptible to this deadlock. One can look for `wait()` functions in `spec091.go` returning true.

`grep -P "^\treturn true" spec091.go  -B 1` should give you an idea.

This PR makes sure no frames are ever sent if we're in the process of shutting down for whatever reason.

I'm not sure if it's tested enough or not. We have had it running for a few days on multiple nodes hammering packages close to line speed while fuzzing the network and restarting RabbitMQ without encountering any deadlocks (after applying #187 and #188). We don't use streadway/amqp to consume thou.

lud4ik:
Thank you for explanation.
Issue [173](https://github.com/streadway/amqp/issues/173) is about the same select in Channel.call.
We tried to solve it by closing rpc or errors chan on channel close for having also closed flag for connection - so on connection shutdown it should close channel, where it should close chan, from which it was waiting in select in call method. We could't find how it could happen than Channel open was invoked without Close on connection shutdown. 
I am happy for you if it works...
But let me show you the hole:
```
// Performs a request/response call for when the message is not NoWait and is
// specified as Synchronous.
func (me *Channel) call(req message, res ...message) error {
	if err := me.send(me, req); err != nil {
		return err
	}
        // what if connection shutdown happen here? 
	if req.wait() {
		select {
		case e := <-me.errors:
			return e

		case msg := <-me.rpc:
			if msg != nil {
				for _, try := range res {
					if reflect.TypeOf(msg) == reflect.TypeOf(try) {
						// *res = *msg
						vres := reflect.ValueOf(try).Elem()
						vmsg := reflect.ValueOf(msg).Elem()
						vres.Set(vmsg)
						return nil
					}
				}
				return ErrCommandInvalid
			} else {
				// RPC channel has been closed without an error, likely due to a hard
				// error on the Connection.  This indicates we have already been
				// shutdown and if were waiting, will have returned from the errors chan.
				return ErrClosed
			}
		}
	}

	return nil
}
```
Your decision solves only if close connection happen before send.

abrander:
Interesting. That would fail if the *client* closes the connection (because `Connection.shutdown()` will be called with a `nil` argument as the error, and therefor never triggering the error handling in `Channel.shutdown()`), but would work as expected when server side closes the connection (Connection.shutdown() will be called like this `me.shutdown(newError(m.ReplyCode, m.ReplyText))`).

We have been testing server side dying/closing/disappearing randomly - therefor never seen this case were the client introduces a race :-/

I think a fix would be in `Channel.shutdown()`. Replace:
```go
		// Notify RPC if we're selecting
		if e != nil {
			me.errors <- e
		}
```

with 

```go
		// Notify RPC if we're selecting
		if e == nil {
			e = ErrClosed
		}

		select {
		case me.errors <- e:
		default:
		}
```

It's untested thou.

lud4ik:
>That would fail if the client closes the connection (because Connection.shutdown() will be called with a nil argument as the error

We tried version where it closed errors chan regardless errors or nil.  I think it could be reproduced playing with Sleep.

lud4ik:
```
select {
        case me.errors <- e:
        default:
        }
```
Default case is redundant, errors chan is bufered and to errors chan is written only in descructor, which is called only once according to struct types

michaelklishin:
I'd extract `atomic.LoadInt32(&me.closed) == 1` into a predicate but otherwise it looks good (and very similar to what other client libraries do).

michaelklishin:
Merged in #223. Thank you!

PR #196, opened 2016-03-28
connection shutdown, channel close data race

imkira:
I am running tests with ```go test -v -race -cpu=1,2,4``` and, I am not sure but, I think I found the following data race (commit ```b4f3ceab0337f013208d31348b578d83c0064744```):

```
==================
WARNING: DATA RACE
Read by goroutine 79:
  github.com/streadway/amqp.(*Channel).call()
      /opt/go/src/github.com/streadway/amqp/channel.go:144 +0x70
  github.com/streadway/amqp.(*Channel).Close()
      /opt/go/src/github.com/streadway/amqp/channel.go:406 +0x207
[ommited]

Previous write by goroutine 80:
  github.com/streadway/amqp.(*Channel).shutdown.func1()
      /opt/go/src/github.com/streadway/amqp/channel.go:104 +0x17b
  sync.(*Once).Do()
      /home/ubuntu/go/src/sync/once.go:44 +0xf6
  github.com/streadway/amqp.(*Channel).shutdown()
      /opt/go/src/github.com/streadway/amqp/channel.go:134 +0x98
  github.com/streadway/amqp.(*Connection).closeChannel()
      /opt/go/src/github.com/streadway/amqp/connection.go:580 +0x38
  github.com/streadway/amqp.(*Connection).shutdown.func1()
      /opt/go/src/github.com/streadway/amqp/connection.go:350 +0x1d5
  sync.(*Once).Do()
      /home/ubuntu/go/src/sync/once.go:44 +0xf6
  github.com/streadway/amqp.(*Connection).shutdown()
      /opt/go/src/github.com/streadway/amqp/connection.go:370 +0x93
  github.com/streadway/amqp.(*Connection).dispatch0()
      /opt/go/src/github.com/streadway/amqp/connection.go:394 +0x4d5
  github.com/streadway/amqp.(*Connection).demux()
      /opt/go/src/github.com/streadway/amqp/connection.go:377 +0x62
  github.com/streadway/amqp.(*Connection).reader()
      /opt/go/src/github.com/streadway/amqp/connection.go:471 +0x322
```

It appears that ```Connection.shutdown``` is accessing ```me.channels``` to close them (goroutine 80) without any kind of mutex locking, while concurrently getting a ```Connection.releaseChannel``` (via ```channel.Close```) which accesses ```me.channels``` too (using the connection lock though).

After trying to fix this by protecting ```Connection.shutdown``` I noticed I started having another race condition due to ```Channel.call``` accessing ```me.send``` while ```Channel.shutdown``` changes it to ```sendClosed```.

Seriously, I don't know if this the correct fix.
I wrote a simple test that causes at least one of these issues to happen, and enabled ```-race``` for travis builds.
There is another data race related to IO handling in ```shared_test.go``` or whatever.
I didn't understand exactly what the problem was, but it looked like a test code-only problem, and since it is out of the scope of this PR I leave it open for you to take a look first.


imkira:
Please notice too that I had to remove the locking around ```Channel.call``` for ```Channel.Confirm``` because I applied the lock while calling ```Channel.send```. Therefore this PR also applies the fix proposed in #187

imkira:
I am sorry, it took me a while to figure it out there was a deadlock in the use of pipes with go1.1. The last commit should work around it.

michaelklishin:
Looks reasonable at first approximation, thank you.

channel.go
```diff
@@ -141,7 +141,11 @@ func (me *Channel) open() error {
 // Performs a request/response call for when the message is not NoWait and is
 // specified as Synchronous.
 func (me *Channel) call(req message, res ...message) error {
-	if err := me.send(me, req); err != nil {
+	me.m.Lock()
```
michaelklishin: There is a separate mutex for sending stuff, `me.sendM`. I think we should use it here.

connection.go
```diff
@@ -578,6 +577,8 @@ func (me *Connection) openChannel() (*Channel, error) {
 // this connection.
 func (me *Connection) closeChannel(ch *Channel, e *Error) {
 	ch.shutdown(e)
+	me.m.Lock()
+	defer me.m.Unlock()
```
michaelklishin: Wouldn't the intent be clearer if we explicitly unlocked after `releaseChannel`? or are we after a particular failure that should be guarded against with defer?

michaelklishin:
Note that this PR also seems to include https://github.com/streadway/amqp/pull/197.

michaelklishin:
Please rebase this against master.

imkira:
> Note that this PR also seems to include #197.

Yes. I left a comment regarding that. Please check:
https://github.com/streadway/amqp/pull/196#issuecomment-202242870

channel.go
imkira: The original code is very misleading.
Please note that m.send starts as `sendOpen` and then switches to `sendClosed`: 

https://github.com/imkira/amqp/blob/0563adcdb2a14308873fce26b14d3637dc2ac685/channel.go#L85
https://github.com/imkira/amqp/blob/0563adcdb2a14308873fce26b14d3637dc2ac685/channel.go#L104

Please, also note that each of those 2 functions uses the mutex you specified for actually sending data.

What I am doing here is to protecting a race condition that could happen if we used me.send directly.
That would collide with the following modification:

https://github.com/imkira/amqp/blob/0563adcdb2a14308873fce26b14d3637dc2ac685/channel.go#L104

connection.go
imkira: Yeah, I don't see any problem with that.
I will change to that then.

imkira:
Just committed one of the changes you asked and the rebase against master.

michaelklishin:
Note that #197, when merged, broke CI in ways that I cannot reproduce in other environments.

channel.go
```diff
@@ -1476,17 +1480,16 @@ exception could occur if the server does not support this method.
...
-
 	if err := me.call(
 		&confirmSelect{Nowait: noWait},
 		&confirmSelectOk{},
 	); err != nil {
 		return err
 	}
 
+	me.m.Lock()
 	me.confirming = true
```
DmitriyMV: Bet advised - the `me.confirming` is also accessed:
https://github.com/streadway/amqp/blob/master/channel.go#L276
https://github.com/streadway/amqp/blob/master/channel.go#L285

Without any sort of mutex.

imkira:
Please note that travis is failing because it is detecting a data race in connection.go that was not fixed by #210 against which I handled the merged conflict by using that version.

https://api.travis-ci.org/jobs/179778933/log.txt?deansi=true

My original branch I believe fixes this by protecting the whole shutdown.
@DmitriyMV @michaelklishin Please let me know what you think before I try to commit it:
https://github.com/imkira/amqp/blob/hotfix/connection-shutdown-race-orig/connection.go#L344

connection.go
```diff
@@ -595,7 +592,9 @@ func (me *Connection) openChannel() (*Channel, error) {
 // this connection.
 func (me *Connection) closeChannel(ch *Channel, e *Error) {
 	ch.shutdown(e)
```
DmitriyMV: You moved mutex from `func (me *Connection) releaseChannel(id uint16)` inside `func (me *Connection) closeChannel(ch *Channel, e *Error)`. And the reason was?

DmitriyMV:
@imkira 
According to Travis CI - it still doesn't builds. I manually applied this pull request and reverted #210. Still - 4 Data Races
https://github.com/DmitriyMV/amqp/commits/pull/196
https://travis-ci.org/DmitriyMV/amqp/builds/179808513
 
Any advice?

shared_test.go
```diff
@@ -78,9 +80,15 @@ func (me *logIO) Write(p []byte) (n int, err error) {
...
 	} else {
-		me.t.Logf("%s close\n", me.prefix)
+		me.logf("%s close\n", me.prefix)
 	}
 	return
 }
+
+func (me *logIO) logf(format string, args ...interface{}) {
+	me.m.Lock()
+	me.t.Logf(format, args...)
```
DmitriyMV: Still fails - according to https://api.travis-ci.org/jobs/179844976/log.txt?deansi=true. On 1,6, but still.

This one is quite allusive, isn't it? 😆

imkira:
@DmitriyMV OK, so I ended up removing all #210 code which was just a subset of the problems I addressed originally with this PR.

The "illusiveness" of the bugs in here are, in my opinion, a reflection of the lack of `-race` for a long time in this project resulting in a huge amount of untested code being committed to the package, tricky code with lack of goroutine synchronization and finalizers, and in general, IMHO, bad (too complex) synchronization practices present in this library :(

I added https://github.com/streadway/amqp/pull/196/commits/524f95a01ca0d2b1ebfd906a027ac2de6e5a0526 to try to fix some of these problems.

With this commit I am synchronizing `Connection.reader()` with `Connection.Close` and doing some other tricky things:
- I found places where the reader and the heartbeater relied on each other but if the heartbeater finished first (via NotifyClose), sometimes reader would be stuck trying to send the heartbeat to a full channel (me.sends).
- I also noticed `me.blocks` were being used in `Connection.dispatch0` without any mutex, despite the fact that they are modifiable via `NofifyBlocked` and accessed during destruction (actually closed, which is even worse).
- I also force connections to be closed in tests so that when the test function finishes processing everything returns back to a clean state.

Please let me know what you think.

By the way, I am getting the following error sometimes on travis https://api.travis-ci.org/jobs/179929218/log.txt?deansi=true:

```
--- FAIL: TestRepeatedChannelExceptionWithPublishAndMaxProcsIssue46 (0.09s)
	shared_test.go:68: issue46 close
	integration_test.go:1411: expected error only on publish, got error on channel.open: Exception (503) Reason: "COMMAND_INVALID - second 'channel.open' seen"
```

Do you have any idea why this may be happening?

DmitriyMV:
Sorry, no. I spent last two days, trying to figure out, why this problem occurs (at random times too). My current thought is that we get channelClose, but don't wait for channelCloseOk in 
```go
func (me *Channel) dispatch(msg message) {
	switch m := msg.(type) {
	case *channelClose:
		me.connection.closeChannel(me, newError(m.ReplyCode, m.ReplyText))
		me.send(me, &channelCloseOk{})
```

I'm not familiar with AMQP protocol, and because of that I'm left to guessing.

michaelklishin:
There is no `channel.close-ok` confirmation. There is `channel.close` which can be sent by both clients and server and there is `channel.close-ok` that confirms it.

DmitriyMV:
@michaelklishin what if we are immediately trying to open channel after we had closed it, with the same channel id? What if server was the one who sent us `channel.close`?

imkira:
@DmitriyMV I thought the reusing of the channel could be a good explanation for what is happening, but when you look at `TestRepeatedChannelExceptionWithPublishAndMaxProcsIssue46`, the function only uses 1 connection, creates 100 channels and sends 10 messages with each serially, and doesn't close any of them. So, I think a bug in the allocator is out of question.

The error message received from the server mentions `second 'channel.open' seen` but, irrespective of that, and according to the reference https://www.rabbitmq.com/amqp-0-9-1-reference.html#constant.channel-error :

```
command-invalid	503	connection	The client sent an invalid sequence of frames, attempting to perform an operation that was considered invalid by the server. This usually implies a programming error in the client.
```

It implies a "invalid sequence of frames". What if we are not waiting enough time between creating a channel and publishing to it? I looked at the code but couldn't find a reason for it, though I found this case block maybe "too inclusive" (possibly triggering `req.Wait()` to succeed when it shouldn't):

https://github.com/streadway/amqp/blob/1b8853833a94be8ba57fd121a8651f7221e5b051/channel.go#L298-L299

DmitriyMV:
But the other tests do not fail, so either the don't generate enough Open/Close channel sequences or it has something to do with situation where channel got closed by rabbitmq due to error.

@imkira I'm open to ideas of how to create a test which will show this situation, without provoking an error from server.

michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

PR #198, opened 2016-03-30
When confirming, return the delivery tag from Publish.

tithonium:
With multiple goroutines publishing messages simultaneously, being able to identify which Publish resulted in which delivery tag is important for processing notifications later. If there's another way to do this, I wasn't able to find it.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -1295,9 +1295,9 @@ When Publish does not return an error and the channel is in confirm mode, the
 internal counter for DeliveryTags with the first confirmation starting at 1.
 
 */
-func (me *Channel) Publish(exchange, key string, mandatory, immediate bool, msg Publishing) error {
+func (me *Channel) Publish(exchange, key string, mandatory, immediate bool, msg Publishing) (uint64, error) {
 	if err := msg.Headers.Validate(); err != nil {
-		return err
+		return 0, err
 	}
 
 	me.m.Lock()
@@ -1325,14 +1325,14 @@ func (me *Channel) Publish(exchange, key string, mandatory, immediate bool, msg 
 			AppId:           msg.AppId,
 		},
 	}); err != nil {
-		return err
+		return 0, err
 	}
 
 	if me.confirming {
-		me.confirms.Publish()
+		return me.confirms.Publish(), nil
 	}
 
-	return nil
+	return 0, nil
 }
 
 /*


imkira:
:+1: 

I also had the same problem, so I had to do my own locking around the channel, but the thing I disliked most was that I had to mimic exactly the same "+1" logic that that already exists in the library. If, for some reason, that changed, or if something didn't get incremented the way I try to mimic, then all will start exploding. 

The only I problem I see with this commit is that it will break semver compatibility with existing applications using this package, but I am totally positive about the need for this.

foolswood:
Watch out for basic.return, which doesn't have a delivery tag, so you can't line them up with ack or publish messages by anything other than channel and receive order as far as I can tell.

imkira:
@foolswood regarding basic.return, I am setting the mandatory flag on the message, and adding a `delivery_tag` with the corresponding value to the headers of the message. Then, the only thing I am serializing, is the publishing of the message. When I receive a basic.return, I check the delivery tag in the headers, and notify the sender about it. Similarly, when the routing succeeds and I get a confirmation, I also notify the sender. Confirmation and returns are handled in parallel. I think this is safe, but it would be better if we could be told by the package what delivery tag is going to be used before it is used. It would be even better if we could control the delivery tags. That way, we could just increment the delivery tag ourselves in a synchronous way, and then handle all publishing, confirmation and returns in parallel.

foolswood:
@imkira I can't see a delivery tag in the amqp frame headers, or the arguments for a return. I've tried reading the spec and looking at the packet raw in wireshark.

My understanding of what a return frame contains is below, hopefully you can show me what I've missed:
```
Headers:
       +-------+---------+--------+
Octet: |   0   |   1:3   |   3:7  |
Field: | Type  | Channel | Length |
Type:  | octet |  short  |  long  |
       +-------+---------+--------+


Payload:
       +-------+--------+------------+------------+----------+-------------+
Field: | Class | Method | Reply-Code | Reply-Text | Exchange | Routing-Key |
Type:  | short | short  |   short    |  shortstr  | shortstr |  shortstr   |
       +-------+--------+------------+------------+----------+-------------+
         (60)     (50)

Frame end (octet)
```

Thanks.

imkira:
@foolswood It is not. I deliberately put the delivery tag (by serializing all publishings to guess the delivery tag), and then building a message with a custom header.

something like

```go
  if msg.Headers == nil {
    msg.Headers = amqp.Table{}
  }
  msg.Headers["delivery_tag"] = int64(deliveryTag)
``` 

Like I said, since `streadway/amqp` does not allow you to get a delivery tag before publishing anything, I am serializing all accesses to this channel, so that I can predict the next delivery tag before calling `channel.Publish`. Then I just put that custom header in the message. When a message gets returned, I know which delivery tag it was when because the whole message is sent back. Because delivery tags are reset to 0 when the channel is closed, I am forcefully timing out all "pending" deliveries so the caller knows something failed.

foolswood:
Fair, I had seen the assumption being made in other libraries and was trying to make sure that nobody else fell into the trap.

michaelklishin:
I'm afraid this is a no-go, at least at the moment: it assumes that publisher confirms are enabled and breaks the API.

PR #200, opened 2016-04-28
Fix data race when closing amqp connection

mouadino:
When using conn.NotifyClose, this later append to a list of channels
``me.closes``, while conn.shutdown() range over this list, so when a
connection is closed, ``conn.NotifyClose()`` may detect the connection closing
in the same time as the ``conn.shutdown()`` call, this is done in two
goroutine, which end up in data race where one appending to me.closes
and the other looping over it.

The fix is to move setting ``noNotify`` to True to the top of
``conn.shutdown()``, this way ``conn.NotifyClose()`` will not append to the
list anymore.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -340,6 +340,10 @@ func (me *Connection) send(f frame) error {
 
 func (me *Connection) shutdown(err *Error) {
 	me.destructor.Do(func() {
+		me.m.Lock()
+		me.noNotify = true
+		me.m.Unlock()
+
 		if err != nil {
 			for _, c := range me.closes {
 				c <- err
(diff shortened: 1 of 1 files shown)


mouadino:
In case tests are needed please let me know.

michaelklishin:
Tests are never a bad idea and force closing connections is something that other client test suites often do, either via `rabbitmqctl` or HTTP API (part of the management plugin).

imkira:
@mouadino how different would be that be from #196 ?
I wonder if that passes the tests I wrote too.

mouadino:
@imkira your PR looks more complete than this one. Any chance we can get your PR merged?

imkira:
@mouadino I hope so. I am waiting for @streadway but he seems to be busy lately.

mouadino:
As mentioned before https://github.com/streadway/amqp/pull/196 will take care of fixing the issue, closing.

PR #207, opened 2016-08-03
Add constants for common amqp exchange types

burdiyan:
It would be useful for library users to deal with constant values instead of raw strings
when choosing the exchange type. It can help to avoid mistakes which are very simple to do with plain string.

For example `net/http` package from the standard library uses the same approach for HTTP statuses and methods and it is pretty useful: https://golang.org/pkg/net/http/#pkg-constants.

### DIFF
--- a/types.go
+++ b/types.go
@@ -11,6 +11,14 @@ import (
 	"time"
 )
 
+// Constants for common AMQP exchange types.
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

PR #210, opened 2016-09-14
fix potential race conditions on closes

rojojo23:
When running our project that uses ampq client with go test -race we are seeing the following race condition reported:

```
==================
WARNING: DATA RACE
Read at 0x00c4202076e8 by goroutine 29:
  vendor/github.com/streadway/amqp.(*Connection).shutdown.func1()
      vendor/github.com/streadway/amqp/connection.go:344 +0x30e

...

Previous write at 0x00c4202076e8 by goroutine 81:
  vendor/github.com/streadway/amqp.(*Connection).NotifyClose()
```

This change simply wraps reading of me.closes with the mutex to avoid that race.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -340,8 +340,12 @@ func (me *Connection) send(f frame) error {
 
 func (me *Connection) shutdown(err *Error) {
 	me.destructor.Do(func() {
+		me.m.Lock()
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

PR #227, opened 2016-11-28
Fix rare deadlock on multiple concurrent conn.Close() calls.

### LINKED ISSUE #226: Concurrent Close calls to Connection struct may result in deadlock.
In rare situations, only first goroutine will successfully get the result inside `func (me *Connection) call`. Others will wait indefinitely on `select`.

DmitriyMV:
Fix #226

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -34,3 +34,20 @@ func TestQueueDeclareOnAClosedConnectionFails(t *testing.T) {
 		log.Fatalf("queue.declare on a closed connection %s is expected to fail", conn)
 	}
 }
+
+func TestConcurrentClose(t *testing.T) {
+	conn, err := Dial("amqp://guest:guest@localhost:5672/")
+	if err != nil {
+		log.Fatalf("Could't connect to amqp server, err = %s", err)
+	}
+
+	for i := 0; i < 5; i++ {
+		t.Run("ConcurrentClose", func(t *testing.T) {
+			t.Parallel()
+			err := conn.Close()
+			if err != nil && err != ErrClosed {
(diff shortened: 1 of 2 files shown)


michaelklishin:
So this seems to fix the same issue as https://github.com/streadway/amqp/pull/196. We'll need some time to weigh the two and merge the preferred version. Thank you.

DmitriyMV:
If I understand #196 correctly - it's about concurrent access from Channels. This fix is about sharing Connection object. Still - I agree with you, and would rather wait for other merge request to pass checks and see how this will affect the problem.

Will investigate again when #196 is green.

michaelklishin:
@DmitriyMV OK, we will take a look at this PR earlier then. I've seen Travis failures that I cannot reproduce locally so it's not clear if the changes in #196 are to blame :(

connection_test.go
```diff
@@ -34,3 +35,23 @@ func TestQueueDeclareOnAClosedConnectionFails(t *testing.T) {
...
 }
+
+func TestConcurrentClose(t *testing.T) {
+	conn, err := Dial("amqp://guest:guest@localhost:5672/")
+	if err != nil {
+		log.Fatalf("Could't connect to amqp server, err = %s", err)
+	}
+
+	wg := sync.WaitGroup{}
+	wg.Add(5)
```
michaelklishin: When I change this value to something higher than 5, I cannot get the suite to pass 20 times in a row, e.g.

```
for i in {1..20}; do go test -v -tags integration_test.go .\/...; done
```

michaelklishin: Never mind, I figured it out.

michaelklishin:
Thank you!

PR #233, opened 2016-12-23
Fix error chan close on channel shutdown.

DmitriyMV:
If errors chan remains open after rabbimq channel shutdown, it could result in deadlock when other goroutine is trying to Close channel again.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -131,6 +131,7 @@ func (ch *Channel) shutdown(e *Error) {
 			ch.confirms.Close()
 		}
 
+		close(ch.errors)
 		ch.noNotify = true
 	})
 }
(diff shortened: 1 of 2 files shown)


michaelklishin:
@DmitriyMV can you please add a test? I assume by "rabbitmq channel shutdown" you mean a channel exception that causes RabbitMQ to close the channel.

DmitriyMV:
Yes, sorry for the incorrect description. In my case this happened when I tried to close the Channel after acknowleding delivery from consumer with autoAck (my incorrect application logic which resulted in channel level exception). Next step is too confirm that it's a reliable way of reproducing this error and writing the proper unit test. Will do.

michaelklishin:
@DmitriyMV thank you for the update.

DmitriyMV:
Sorry - my initial assumption was incorrect. The deadlock happens when you try to concurrently close channel and it's connection. You may require several attempts to reproduce it (usually 2-3 times).

The commit above fixes it.

DmitriyMV:
Is there anything that stops from merging this branch?

michaelklishin:
@DmitriyMV it needs QA. Chances are it may not happen until January 3rd or so.

michaelklishin:
Merging per discussion with @streadway and @gerhard.

Issue #202, opened 2016-05-15
Add URI validation


tleyden:
I ran into [this snag](https://groups.google.com/d/msg/rabbitmq-users/6E6riP8IvCw/vV8Ko0LTCQAJ) when connecting, which turned out to be human error when passing the connection URI.

Would it be possible to add validation to make this a bit more straightforward to debug?

Eg, throw an error on presumably invalid connection URI's like `amqp://admin:PASSWORD@rabbitmq-service/ -http_port=8080`?

michaelklishin:
All 3 maintainers agree this is a very good idea. Hopefully there won't be any need for breaking public API changes.

michaelklishin:
Note to self: the preferred way of URI validation seems to be [url.ParseRequestURI](https://github.com/asaskevich/govalidator/blob/fdf19785fd3558d619ef81212f5edf1d6c2a5911/validator.go#L84).

gerhard:
@tleyden updating to latest will fix this.

PR #241, opened 2017-02-11
Validate URI before attempting to use it

michaelklishin:
References #202, whether this validation is sufficient is up for debate.


michaelklishin:
@gerhard no objections.

PR #244, opened 2017-02-24
rabbimq uri can not contains symbol (?)

davidstack:
if my mq uri is:amqp://docker:123456a?@10.110.20.179:5672/ ,it will failed when connect to mq with this log:
AMQP scheme must be either 'amqp://' or 'amqps://'

when i check the code,i find the bug:
`func ParseURI(uri string) (URI, error) {
builder := defaultURI

u, err := url.Parse(uri)
if err != nil {
	return builder, err
}

defaultPort, okScheme := schemePorts[u.Scheme]

if okScheme {
	builder.Scheme = u.Scheme
} else {
	return builder, errURIScheme
}`
this function use net.url to parase the amq uri. we know that in the url,? is used to set parament.

this commit can fix this bug

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


davidstack:
because i add a regex check for this uri:^amqp(s)?://[^:\/\/]+:[^:\/\/]+@[^:\/\/]+:[0-9]+\/[^:\/\/]*
can this uri test be removed.i do not think this is useful uri. 
`{
		url:      "amqp://",
		username: defaultURI.Username,
		password: defaultURI.Password,
		host:     defaultURI.Host,
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://localhost/",
	},

	{
		url:      "amqp://:@/",
		username: "",
		password: "",
		host:     defaultURI.Host,
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://:@localhost/",
	},

	{
		url:      "amqp://user@",
		username: "user",
		password: defaultURI.Password,
		host:     defaultURI.Host,
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://user@localhost/",
	},

	{
		url:      "amqp://user:pass@",
		username: "user",
		password: "pass",
		host:     defaultURI.Host,
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://user:pass@localhost/",
	},

	{
		url:      "amqp://guest:pass@",
		username: "guest",
		password: "pass",
		host:     defaultURI.Host,
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://guest:pass@localhost/",
	},

	{
		url:      "amqp://host",
		username: defaultURI.Username,
		password: defaultURI.Password,
		host:     "host",
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://host/",
	},

	{
		url:      "amqp://:10000",
		username: defaultURI.Username,
		password: defaultURI.Password,
		host:     defaultURI.Host,
		port:     10000,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://localhost:10000/",
	},

	{
		url:      "amqp:///vhost",
		username: defaultURI.Username,
		password: defaultURI.Password,
		host:     defaultURI.Host,
		port:     defaultURI.Port,
		vhost:    "vhost",
		canon:    "amqp://localhost/vhost",
	},

	{
		url:      "amqp://host/",
		username: defaultURI.Username,
		password: defaultURI.Password,
		host:     "host",
		port:     defaultURI.Port,
		vhost:    defaultURI.Vhost,
		canon:    "amqp://host/",
	},

	{
		url:      "amqp://host/%2F",
		username: defaultURI.Username,
		password: defaultURI.Password,
		host:     "host",
		port:     defaultURI.Port,
		vhost:    "/",
		canon:    "amqp://host/",
	},`

michaelklishin:
Thank you but we will go with #241 instead of using regular expressions. Question marks in URIs have a special meaning and thus must be %-encoded.

davidstack:
@michaelklishin  ok.i understand. but i still think amqp uri is not the net url.  when i use the java client and python client of the amqp, they don't have this problem.

michaelklishin:
@davidstack an AMQP URI may or may not be a URL but it is certainly a URI. FWIW the state of the art in URI parsing in Go seems to be `ParseRequestURI`, at least I could not find a better option and that's what a [popular validation library uses](https://github.com/asaskevich/govalidator/blob/master/validator.go#L84)

I disagree that `amqp://host/%2F` is not a "useful URI". It is certainly used by a ton of tests that are connecting to a localhost node and perfectly fine with using default credentials. Same goes for `amqp:///vhost` and `amqp://host/` (the latter is [incorrect](https://www.rabbitmq.com/uri-spec.html) but it is still a value used — and an extremely common mistake in client libraries)

gerhard:
@davidstack I think that the simplest solution for `amqp://docker:123456a?@10.110.20.179:5672/` is to not use `?` in the password. You can replace the problematic question mark with `!` or even 😄  - not that I would recommend the latter. Your password will still contain 3 types of characters and therefore not be any less secure against brute force attacks, and we don't need to add any new code - everyone wins 😄

PR #246, opened 2017-02-25
Don't get stuck on mixed sequences of multiple and single confirmations with unordered delivery tags

zenovich:
Confirmations of publishings get stuck sometimes on mixed sequences of multiple and single confirmations with unordered delivery tags.

We just need to call resequense() for multiple publisher confirmations to fix the issue (the test added contains a real-life sequence of confirmations)

### DIFF
--- a/confirms.go
+++ b/confirms.go
@@ -78,6 +78,7 @@ func (c *confirms) Multiple(confirmed Confirmation) {
 	for c.expecting <= confirmed.DeliveryTag {
 		c.confirm(Confirmation{c.expecting, confirmed.Ack})
 	}
+	c.resequence()
 }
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you! We will take a look next week.

michaelklishin:
The test looks reasonable, as is the fix.

michaelklishin:
@zenovich thank you! It must have been a pretty tricky issue to reproduce and debug.

zenovich:
@michaelklishin At that moment, I was tired of using workarounds such as reconnections on confirmation timeouts I had been trying for months. That's why I had started searching for the root issue. Surprisingly, it took me just several hours to find the bug and create a patch.

Thanks for approving and merging! It'll help all people who use publishing confirmations in amqp.

PR #248, opened 2017-03-02
Only use TLS when the connection scheme is “amqps://“.

### LINKED ISSUE #190: Unable to connect to amqp:// via DialTLS
The [documentation for `DialTLS`](https://godoc.org/github.com/streadway/amqp#DialTLS) says:

> DialTLS uses the provided tls.Config when encountering an amqps:// scheme.

This sounds great for us, since we prefer SSL/TLS in production but typically develop without it. I expected I could pass some reasonable `tls.Config` regardless of the connection string:

``` go
package main

import (
    "crypto/tls"
    "testing"

    "github.com/streadway/amqp"
)

func TestDialTLSWithAMQP(t *testing.T) {
    uri := "amqp://user:pass@rabbitmq-1.lxc/test"

    conn, err := amqp.DialTLS(uri, &tls.Config{MinVersion: tls.VersionTLS12})
    if err != nil {
        t.Fatalf("Expected no error, got %v", err)
    }
    defer conn.Close()
}
```

However, using `DialTLS` this way results in an error:

```
=== RUN TestDialTLSWithAMQP
--- FAIL: TestDialTLSWithAMQP (0.16s)
        cxn_test.go:15: Expected no error, got tls: oversized record received with length 20480
FAIL
exit status 1
FAIL    test    0.164s
```

Changing the connection string to `amqps://user:pass@rabbitmq-1.lxc/test` works perfectly.

Am I doing something wrong? Is the documentation incorrect in this case?

For now we do the following:

``` go
var conn *amqp.Connection
var err error

if strings.HasPrefix(uri, "amqps") {
    conn, err = amqp.DialTLS(uri, &tls.Config{MinVersion: tls.VersionTLS12})
} else {
    conn, err = amqp.Dial(uri)
}
```

domodwyer:
Change the default dialer to only establish a TLS session if the connection scheme is `amqps`.

When running the integration tests locally I routinely get a failure for a single test:
```
=== RUN   TestExchangeDeclarePrecondition
--- FAIL: TestExchangeDeclarePrecondition (33.55s)
	integration_test.go:1759: dial integration server: Exception (501) Reason: "read tcp 127.0.0.1:51262->127.0.0.1:5672: i/o timeout"
```

However running this test in isolation (`-run TestExchangeDeclarePrecondition`) always passes - not sure what the craic is here.

Fixes #190.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -165,10 +165,6 @@ func DialConfig(url string, config Config) (*Connection, error) {
 		config.Vhost = uri.Vhost
 	}
 
-	if uri.Scheme == "amqps" && config.TLSClientConfig == nil {
-		config.TLSClientConfig = new(tls.Config)
-	}
-
 	addr := net.JoinHostPort(uri.Host, strconv.FormatInt(int64(uri.Port), 10))
 
 	dialer := config.Dial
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you!

PR #250, opened 2017-03-13
Avoid leaking channel allocations in Connection.openChannel()

domodwyer:
When opening a new channel (using `Connection.Channel()`), a new channel is allocated and added to the connection channel map (in `Connection.allocateChannel()`) before sending a `channel.open` frame to the server (in `Connection.openChannel()`).

If the `channel.open` frame fails, the allocated channel is not released, leading to a memory leak.

This patch ensures that the allocated channel is released if the `channel.open` frame send fails.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -605,6 +605,7 @@ func (c *Connection) openChannel() (*Channel, error) {
 	}
 
 	if err := ch.open(); err != nil {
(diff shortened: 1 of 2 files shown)


michaelklishin:
Good catch, thank you!

domodwyer:
Thanks for the speedy merge!

Issue #237, opened 2017-01-27
Setting client Locale


AndrewWPhillips:
I am trying to connect to a server that does not like an empty Locale string. As far as I can tell Connection.openTune in connection.go uses a hard-coded Locale of "". Is there any way to set the Locale string to say "en-US"?

Sorry if I am missing something but I am new to GO.

gerhard:
Have you tried running the producer & consumer examples? What errors are you getting? Which version of RabbitMQ server are you running?

I am running the producer example against RabbitMQ 3.6.6, as installed by `brew install rabbitmq`

```
go run _examples/simple-producer/producer.go

2017/02/03 12:02:18 dialing "amqp://guest:guest@localhost:5672/"
2017/02/03 12:02:18 got Connection, getting Channel
2017/02/03 12:02:18 got Channel, declaring "direct" Exchange ("test-exchange")
2017/02/03 12:02:18 enabling publishing confirms.
2017/02/03 12:02:18 declared Exchange, publishing 6B body ("foobar")
2017/02/03 12:02:18 waiting for confirmation of one publishing
2017/02/03 12:02:18 confirmed delivery with delivery tag: 1
2017/02/03 12:02:18 published 6B OK
```

gerhard:
FWIW, RabbitMQ does not use this value. We could make it non-blank and set locale to `"en-us"`. Would that help?

AndrewWPhillips:
Thanks gerhard for the reply.  I am not talking to RabbitMQ but a custom AMQP 0.9.1 server. I could possibly get the server code modified to assume "en-US" if locale is not set (the server locale check is just a sanity check), but having the library default to "en-us" would be a solution.

However, there are a few other things that I have since encountered so I might just have to fork the streadway/AMQP library anyway.

Andrew (aphillips801 AT gmail DOT com)

RichardKnop:
@gerhard Perhaps this could be blank by default but there should be some way to override the default blank value.

michaelklishin:
We will set the default (and possibly include the `en_US` locale if it's not provided).

gerhard:
We are agreeing to always set en_US as the default locale. [As per AMQP 0-9-1 spec, the server MUST support en_US](https://www.rabbitmq.com/amqp-0-9-1-reference.html).

PR #252, opened 2017-03-17
Default to en_US locale

gerhard:
Default to en_US, [as per AMQP 0-9-1 spec](https://www.rabbitmq.com/amqp-0-9-1-reference.html).

Some AMQP servers, such as the one mentioned in #237, require a locale to be set. Since I am not aware of any AMQP server using this locale for anything else other than satisfying the AMQP spec, defaulting to en_US seems a sensible.


gerhard:
Is it OK to merge this @streadway ?

PR #269, opened 2017-05-17
Protect allocator

cesar0094:
Race condition with `connection.go:418`.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -814,7 +814,9 @@ func (c *Connection) openComplete() error {
 		_ = deadliner.SetDeadline(time.Time{})
 	}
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
@cesar0094 thank you but this is a bit light on details. What exactly is the shared resource we are guarding? `newAllocator` seems to return a new record and only called during connection negotiation, so what is the shared resource (or critical section) we are trying to ensure exclusive access to?

We would like to have a test that reproduces the problem reasonably frequently.

michaelklishin:
Closing as we do not have a way to reproducing this supposed race condition and `newAllocator` seems to be safe as is.

PR #276, opened 2017-07-06
Fix release channel id before send channelCloseOk cause CHANNEL_ERROR

heimonsy:
When Server sends a ChannelClose with some exception(like ACCESS_REFUSED), and the user opens a new channel on the same connection immediately. The new channel will use the old one's channel id, and it will cause a new exception `Exception (504) Reason: "CHANNEL_ERROR - second 'channel.open' seen"` which closed the connection.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -276,6 +276,7 @@ func (ch *Channel) dispatch(msg message) {
 	case *channelClose:
 		ch.connection.closeChannel(ch, newError(m.ReplyCode, m.ReplyText))
 		ch.send(&channelCloseOk{})
+		ch.connection.releaseChannel(ch.id)
 
 	case *channelFlow:
 		ch.notifyM.RLock()
@@ -418,6 +419,7 @@ It is safe to call this method multiple times.
 
 */
 func (ch *Channel) Close() error {
+	defer ch.connection.releaseChannel(ch.id)
 	defer ch.connection.closeChannel(ch, nil)
 	return ch.call(
 		&channelClose{ReplyCode: replySuccess},
--- a/connection.go
+++ b/connection.go
@@ -633,7 +633,6 @@ func (c *Connection) openChannel() (*Channel, error) {
 // this connection.
 func (c *Connection) closeChannel(ch *Channel, e *Error) {
 	ch.shutdown(e)
-	c.releaseChannel(ch.id)
 }
 
 /*


channel.go
```diff
@@ -276,6 +276,7 @@ func (ch *Channel) dispatch(msg message) {
 	case *channelClose:
 		ch.connection.closeChannel(ch, newError(m.ReplyCode, m.ReplyText))
 		ch.send(&channelCloseOk{})
+		ch.connection.releaseChannel(ch.id)
```
michaelklishin: Here we release channel ID after sending a `channel.close-ok`, which [causes channel process termination on the RabbitMQ end](https://github.com/rabbitmq/rabbitmq-server/blob/stable/src/rabbit_channel.erl#L879).

This is not consistent with what we do when the client initiates channel closure below.

channel.go
```diff
@@ -418,6 +419,7 @@ It is safe to call this method multiple times.
 
 */
 func (ch *Channel) Close() error {
+	defer ch.connection.releaseChannel(ch.id)
```
michaelklishin: Should we release channel ID only after sending a `channel.close` and receiving a `channel.close-ok`? [This channel.close handler comment](https://github.com/rabbitmq/rabbitmq-server/blob/stable/src/rabbit_channel.erl#L890) to me suggests that before a client receives a `channel.close-ok`, it is not supposed to reuse the channel ID. Even if that's not the case, by releasing at the latest possible moment we reduce the probability of races for workloads with channel churn.

michaelklishin:
Thank you! This needs a bit of extra consideration.

channel.go
heimonsy: Sorry, forgive my poor English. I didn't get your point. Here I use a defer, I'm sure it will release after connection received the `channel.close-ok`, it will prevent  `channel.open` for the same channel number.

michaelklishin: You are right. I'm going to see if we can achieve the same effect without using `defer` (nothing wrong with it, just easier to reason about things without delayed operations).

connection.go
```diff
@@ -633,7 +633,6 @@ func (c *Connection) openChannel() (*Channel, error) {
 // this connection.
 func (c *Connection) closeChannel(ch *Channel, e *Error) {
 	ch.shutdown(e)
-	c.releaseChannel(ch.id)
```
michaelklishin: If we do this documentation for a few methods needs updating. I'm looking into if we can adopted.

michaelklishin:
@heimonsy thanks, you have uncovered a couple of subtle issues related to server-initiated `channel.close` in the library. I hope they are both addressed in master. Please give it a try.

heimonsy:
Thank you for fixed this problem. It can make my testing pass now.  😸

PR #277, opened 2017-07-06
Fix publish channel redial

dimm0:
When publish connection is closed, it goes to infinite loop. We need to watch the "confirmed" channel to be closed and break the loop to redial.


michaelklishin:
Thank you. Wow, this example is becoming a blueprint for a production service 👍

PR #286, opened 2017-08-21
Fix handling of IPv6 hostnames

jleroy:
The current implementation of host/port parsing expect that IPv6 hosts are always enclosed in square brackets, which is incorrect (an IPv6 address must be enclosed with brackets **only** when followed by a port), and break on some IPv6 addresses. 

Example using the consumer example included in this repository:

```
# ./consumer 
2017/08/21 06:46:27 dialing "amqp://guest:guest@[fe80::f816:3eff:fe63:7434]:5672/"
2017/08/21 06:46:27 Dial: dial tcp [fe80::f816:3eff:fe63:7434]:5672: connect: invalid argument
```

I've fixed this by rewriting the `ParseURI()` function using the `Parse()` function provided by the `net/url` package.

This patch is used in production since more than tree months without any error.


michaelklishin:
Thank you, relying on `net/url` more is definitely a good idea. So is submitting a PR after a few months of testing 👍

michaelklishin:
Note: this breaks Go 1.6 and 1.7 compatibility but @streadway, @gerhard and I decided to support two most recent releases only several months ago, which currently means 1.8 and 1.9. I'm going to document this requirement separately.

Issue #278, opened 2017-07-13
Channels do not have a per-operation timeout


pierrre:
Hi,

I think I've found a bug with `Connection.Channel()`: sometimes the call is blocked indefinitely.

My scenario:
I've got a queue in a vhost. I start to consume on this queue.
Then I delete the vhost from the RabbitMQ web interface. Obviously, the channel is closed.

In my code, I retry in an infinite loop to open a new  channel, then to consume on this channel.
Consume() returns an error:
`Exception (403) Reason: "ACCESS_REFUSED - access to queue 'xxxx' in vhost 'xxxx' refused for user 'guest'"` (it's expected)
After a few seconds, my program is blocked indefinitely.

I think that the RabbitMQ server has closed the connection from its side, because I don't see the connection in the web interface anymore.
Nothing relevant in the logs.

Here is the stack trace for the goroutine:
```
goroutine 34 [select]:
github.com/streadway/amqp.(*Channel).call(0xc42025b8c0, 0xa43020, 0xc4202690b0, 0xc42044bd70, 0x1, 0x1, 0x1, 0xc4203074a0)
	github.com/streadway/amqp/channel.go:176 +0x198
github.com/streadway/amqp.(*Channel).open(0xc42025b8c0, 0xc42025b8c0, 0x0)
	github.com/streadway/amqp/channel.go:165 +0xb3
github.com/streadway/amqp.(*Connection).openChannel(0xc420295a40, 0xc420024600, 0x77e7a0, 0x8a2480)
	github.com/streadway/amqp/connection.go:624 +0x55
github.com/streadway/amqp.(*Connection).Channel(0xc420295a40, 0xa42820, 0xc42033b4c0, 0xc420295a40)
	github.com/streadway/amqp/connection.go:646 +0x2b
```

michaelklishin:
RabbitMQ 3.6.x will NOT close connections to a vhost that is deleted (3.7.0 will). So I'd expect something similar since the client isn't notified unless it attempts to perform an operation.
How exactly it would run into an error remains to be seen.

Please post your code or a small snippet that can be used to reproduce.
A [traffic capture](https://www.rabbitmq.com/amqp-wireshark.html) would help immensely.

This should be a non-issue with 3.7.0 since clients will observe a server-initiated connection closure.

michaelklishin:
The client is blocked in a select from `<-ch.errors`, which in this case will not get any notifications.

``` go
func (ch *Channel) call(req message, res ...message) error {
[...]

PR #296, opened 2017-10-10
Add timeout for channel rpc operations

ponumapathy:
Fix for the issue - https://github.com/streadway/amqp/issues/278

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
+	"time"
 )
 
 // 0      1         3             7                  size+7 size+8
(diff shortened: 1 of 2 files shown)


channel.go
```diff
@@ -197,6 +198,8 @@ func (ch *Channel) call(req message, res ...message) error {
 			// error on the Connection.  This indicates we have already been
 			// shutdown and if were waiting, will have returned from the errors chan.
 			return ErrClosed
+		case <-time.After(ch.connection.Config.Heartbeat):
+			return ErrChannelOpTimeout
```
ponumapathy: is there any cleanup there might be required here? to handle rpc responses delivered after this select loop exits due to a timeout?

ponumapathy:
closed since i think this needs more discussion - see referenced issue.

PR #304, opened 2017-11-21
Channel RPC operations timeout

ponumapathy:
Fix for - https://github.com/streadway/amqp/issues/278

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


michaelklishin:
@streadway does this look good to you? Is there a timer leak I'm overlooking? :)

michaelklishin:
This change comes with a certain penalty because we acquire a lock and in general do more work on the hot path but not having a timeout on channel operation is a worse thing to have.

PR #315, opened 2018-01-26
Open Vhost failed should return the real error which send from server.

barryz:
In the latest version of RabbitMQ,  I set the `max-connections` limitation for a vhost. When I opening
several connections that try to connection to RabbitMQ.  The client shows an `ErrVhost` error. However, we expect to receive a similar error where send from server like this: 
```bash
Exception (530) Reason: "NOT_ALLOWED - access to vhost 'test_vhost' refused for user 'guest': connection limit (2) is reached"
```
Therefore, I think it is reasonable to return the original error where send from server.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -326,8 +326,8 @@ func TestOpenFailedVhost(t *testing.T) {
 	}()
 
 	c, err := Open(rwc, defaultConfig())
-	if err != ErrVhost {
-		t.Fatalf("expected ErrVhost got: %+v on %+v", err, c)
+	if err == nil {
+		t.Fatalf("expected not nil error got: %+v on %+v", err, c)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
This is a breaking API change and this client is ridiculously conservative about those. However, there is only code path for opening a connection. Let's wait and see what others think.

streadway:
I've also been bitten by debugging this error. The error API for connection handshake doesn't easily carry the stage of the handshake along with the reason when using package level errors.

There are a few uses of comparing ErrVhost in the wild: https://github.com/search?q=ErrVhost If the error protocol were changed - would you be willing to create PRs to on the repositories affected?

Alternatively, for usability, what do you think about improving the error message to include max-connection possibility and to investigate the server logs for the actual error?

michaelklishin:
The vast majority of the `ErrVhost` uses are vendored code. I haven't inspected all 25 pages but looks like there may be a few dozens of projects that would need updating.

streadway:
Ok, until we upgrade the packages, let's include this as a possible reason to the error string of the ErrVhost variable.

michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

PR #316, opened 2018-01-31
Revert "Channel RPC operations timeout"

michaelklishin:
Reverts streadway/amqp#304


michaelklishin:
Reverting per https://github.com/streadway/amqp/issues/278#issuecomment-361838086.

Issue #318, opened 2018-02-27
No lock in ack/nack


svarogg:
Is there a reason why there's no call for ch.m.lock() in .Ack() and .Nack(), the same way there's in .Publish() etc?

This causes .Ack() and .Nack() to be non-thread safe?

michaelklishin:
Probably just a an oversight. Both `basic.publish` and `basic.{ack,nack}` have a lot of similarities in the protocol (e.g. they don't have a response), and so are their concurrency "characteristics". Please submit a PR and we will discuss.

PR #321, opened 2018-03-08
Added locking to channel.Ack and channel.Nack to make them both goroutine-safe

svarogg:
This should take care of #318


michaelklishin:
Thank you!

Issue #306, opened 2017-11-22
Consumer tags exceeding maximum allowed length are generated


deadok22:
A consumer tag is generated when `Channel.Consume` is called with `consumer` set to empty string. The generated value may exceed maximum allowed length resulting in a [failure to map](https://github.com/streadway/amqp/blob/ff791c2d22d3f1588b4e2cc71a9fba5e1da90654/consumers.go#L122) `basicDeliver` messages back to consumers and ultimately [dropping messages without acks or nacks](https://github.com/streadway/amqp/blob/ff791c2d22d3f1588b4e2cc71a9fba5e1da90654/channel.go#L328).

Change the tag generator so that it never generates names exceeding maximum allowed length. See #305 for similar problems.

I didn't run integration tests yet, but I plan to submit a pull request with the fix: https://github.com/deadok22/amqp/commit/440a0930e2eaecd094302ce250a636aace65d8c4

michaelklishin:
Please submit a PR.

deadok22:
Pull request: #322

PR #322, opened 2018-03-09
#306 Always generate a correct consumer tag

deadok22:
Consumer tags exceeding maximum length have resulted in failures to map
basicConsume messages back to consumers, ultimately dropping deliveries
without sending ack or nack.

A proper fix is to add validation for all strings in configuration and
fail early if any of the configured parameters exceeds limits set by
the protocol.

### DIFF
--- a/consumers.go
+++ b/consumers.go
@@ -6,16 +6,30 @@
 package amqp
 
 import (
-	"fmt"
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you!

PR #329, opened 2018-04-26
Add option to force disable heartbeats

hpk:
This adds a `DisableHeartbeat bool` option in `Config` that overrides the `pick` behavior to set the heartbeat interval to 0, even if the server proposes one.

Justification: Our client is running in an AWS Lambda, which can have its execution context "frozen" when the Lambda is idle. This means that if the Lambda is invoked intermittently, there's no guarantee that heartbeats will fire in time. We're also using CloudAMQP, which doesn't provide any control over its server-proposed heartbeat intervals.

As a side note, CloudAMQP recommends disabling heartbeats when using their service, since they implement TCP keep-alive: https://www.cloudamqp.com/docs/celery.html (See the note under "broker_heartbeat")

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -46,6 +46,9 @@ type Config struct {
 	FrameSize  int           // 0 max bytes means unlimited
 	Heartbeat  time.Duration // less than 1s uses the server's interval
 
+	// Force heartbeats to be disabled, even if server specifies an interval
+	DisableHeartbeat bool
+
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you for the contribution.

All other clients use the same negotiation logic that's found in `pick` and CloudAMQP have been OK with it for years. You can set heartbeat timeout to a really high value (say, 30 minutes) which effectively disables them.

Issue #63, opened 2013-05-22
Enable mechanisms for confirmed delivery: publisher confirms and transactions


yadvr:
I've not explored this amqp library in details but the idea is to have RabbitMQ's extension -- publisher confirms over a channel. Also, implement transactions if not done already.
I started by writing my library but found this one, so it makes sense to learn the work that has been done in it till now and contribute so that it can be used in production env.

Sean, thanks for your library. Can you please list features which have been missing so we know on the roadmap what needs to be done or needs to be fixed? My requirement specifically is to have a golang library which supports some way of transactions and provides guarantee of communication over full of latency geo distributed brokers and consumers.

PR #336, opened 2018-06-06
Channel exposes confirms in order to get sequential number of the published messages

digitalcrab:
Hello guys.

It looks like the current implementation doesn't allow the client to get DeliveryTag in the moment of publishing. It makes life a bit harder, and the user should implement his way of tracking published messages. Would it be better to expose counter which already in place?

Related to #63 question

Thank you

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -1585,3 +1585,8 @@ func (ch *Channel) Reject(tag uint64, requeue bool) error {
 		Requeue:     requeue,
 	})
 }
+
+// Confirms returns control structure over confirm mode
+func (ch *Channel) Confirms() *confirms {
+	return ch.confirms
+}
--- a/confirms.go
+++ b/confirms.go
@@ -20,6 +20,14 @@ func newConfirms() *confirms {
 	}
 }
 
+// Published returns sequential number of published messages
+func (c *confirms) Published() uint64 {
+	c.m.Lock()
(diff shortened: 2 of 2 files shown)


michaelklishin:
You are right, and other clients usually do it.

streadway:
@michaelklishin woops this was merged without discussing the impact of exposing the other package internal bookeeping methods from the `Confirms` struct like `One`. I think we should revert this commit as soon as possible before it reaches too many call sites.

This API design requires a mutex around both `Publish` and `Confirms.Published` to be concurrent safe. This is going to be very easy to get wrong.

Also, if a client calls any of these methods: https://godoc.org/github.com/streadway/amqp#Confirms the internal state of publisher confirms will get out of sync.

Let's look again at exposing the sequential delivery tag without exposing this internal type.

@digitalcrab would it be sufficient if there was a publish method that returned the delivery tag after publish?

michaelklishin:
Reverted.

What would a publish function return if confirms were not enabled on the channel, a `nil`?

Zenimax-DSpeed:
We pulled this library at the point in time where this was in, and built around this issue.
Reverting is a breaking library change for us for functionality that we've built.

I agree that the implementation was bad, but maybe on a library like this we could have semantic / module friendly versioning at some point?

Issue #333, opened 2018-05-21
Deadlock when in confirm mode.


xReprisal:
I'm running into deadlock on producer when channel is in confirm mode.

I think it has to do with `confirms.go`  mutex around [published](https://github.com/streadway/amqp/blob/master/confirms.go#L31) property. When I alter this property to be `atomic.Uint64` using go.uber.org/atomic package issue is solved. 

```
package amqp

import (
	"sync"
	"fmt"
	"go.uber.org/atomic"
)

// confirms resequences and notifies one or multiple publisher confirmation listeners
type confirms struct {
	m         sync.Mutex
	listeners []chan Confirmation
	sequencer map[uint64]Confirmation
	published *atomic.Uint64
	expecting uint64
}

// newConfirms allocates a confirms
func newConfirms() *confirms {
	return &confirms{
		sequencer: map[uint64]Confirmation{},
		published: atomic.NewUint64(0),
		expecting: 1,
	}
}
...
// publish increments the publishing counter
func (c *confirms) Publish() uint64 {
	c.published.Inc()
	return c.published.Load()
}
```

I can create test if this does not convey the problem clearly.

michaelklishin:
@xReprisal that code does acquire a lock but we'd consider a PR that switches to an atomic data type. A test that runs a reasonably long time and allows to reproduce would be appreciated. Thanks.

xReprisal:
I think I've finally realized my issue. I was doing something along these lines:
```

func (p *Producer) run() {
	for {
		select {
		case <-p.confirm:
			// logic goes here
		case msg := <-p.publish:
			if err := p.ch.Publish(p.exchangeName, p.routingKey, true, false, msg); err != nil {
				// logic goes here
			}
		}
	}
}

func (p *Producer) Publish(msg samqp.Publishing) {
	p.publish <- msg
}
```

Seems like the problem was with listening to both `p.publish` and `p.confirm` on the same select statement. When I moved `p.confirm` outside there are no deadlocks.

PR #337, opened 2018-06-12
Use atomic data type to count published messages

xReprisal:
When running publisher in confirm mode I kept running into deadlocks. After some digging I have figured out that mutex on publish is the culprit. Atomic value fixes the issue by removing mutex from the function.

Issue: https://github.com/streadway/amqp/issues/333

### DIFF
--- a/confirms.go
+++ b/confirms.go
@@ -1,23 +1,27 @@
 package amqp
 
-import "sync"
+import (
+	"sync"
+	"sync/atomic"
+)
 
 // confirms resequences and notifies one or multiple publisher confirmation listeners
 type confirms struct {
 	m         sync.Mutex
 	listeners []chan Confirmation
 	sequencer map[uint64]Confirmation
-	published uint64
+	published atomic.Value
 	expecting uint64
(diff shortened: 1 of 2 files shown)


confirms.go
```diff
@@ -29,11 +33,8 @@ func (c *confirms) Listen(l chan Confirmation) {
 
 // publish increments the publishing counter
 func (c *confirms) Publish() uint64 {
-	c.m.Lock()
-	defer c.m.Unlock()
-
-	c.published++
-	return c.published
+	c.published.Store(c.published.Load().(uint64) + 1)
```
gaaf: This seems racy. What prevents another process from altering c.published between Load and store?

Instead, IMHO, just using atomic.AddUint64(&c.published) to increment c.published and removing the locking should be all that's necessary to accomplish lock-less incrementing.

xReprisal: I didn't know that its possible to do `atomic.AddUint64(&c.published)`. I'll update my PR. 
Thanks for the insight!

PR #345, opened 2018-09-07
Add go v1.11 to supported versions, soft drop go v1.9 (n-1 + tip project policy)

nodo:
Also, update README to reflect go version tested in travis.

PR #362, opened 2018-10-28
Bump Go versions

HaraldNordgren:


gerhard:
Thank you, now supported via #345

PR #380, opened 2019-01-06
Fix #363 where int is not treated as valid field

### LINKED ISSUE #363: int as Table value / validation error
Hi,

I tried to use an `int` as header in an amqp message via the `Table`type. 
This resulted in a validation error claiming the type `int` isn't supported, at least when the value was `0`. A look into the `types.go@validateField` function `int` isn't in the list, but `int16/32/64` are. 

As far as I understood `int` in go, it's at least equivalent to `int32`, also it isn't  an alias. But from the standpoint of mapping an integer datatype into the amqp message, I don't understand why it should be banned. 

Is this an error or an active decision, if the later can you elaborate this decision (and maybe add a comment to the `Table` type)

Thanks!

asahaf:
Fixes #363.

### DIFF
--- a/write.go
+++ b/write.go
@@ -308,6 +308,11 @@ func writeField(w io.Writer, value interface{}) (err error) {
 		binary.BigEndian.PutUint16(buf[1:3], uint16(v))
 		enc = buf[:3]
 
+	case int:
+		buf[0] = 'I'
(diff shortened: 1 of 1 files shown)


michaelklishin:
Good catch, thank you.

PR #385, opened 2019-02-14
Allow to use default dial with custom connection timeout

amberovsky:


michaelklishin:
This makes sense to me 👍

.gitignore
```diff
@@ -2,3 +2,6 @@ certs/*
 spec/spec
 examples/simple-consumer/simple-consumer
 examples/simple-producer/simple-producer
+
+# GoLand IDE project files
+/.idea
```
tania-paddle: this should be in your global .gitignore

michaelklishin: I made the ignore to be a bit more fine-grained and follow JetBrains' guidelines. It makes sense to me to ignore project-specific files in a project-specific `.gitignore`.

Issue #266, opened 2017-05-11
Implement context


marwan-at-work:
I imagine this is an API-breaking change to the amqp package. But is this something the maintainers have thought about? Is it possible? 

This is my scenario: 

I send 3 messages to 3 workers. If one worker fails, I'd like to tell the other workers that you can stop working because one of you failed. 

General motivation: 

the context package is becoming a common pattern in clients, like the docker client, github-api client, and net/http package.

michaelklishin:
This was discussed before. Currently this client does not use branching, tags or any kind of versioning so it will have to be a separate package or something like that.

marwan-at-work:
@michaelklishin maybe we could use gopkg.in to create v2? also, i'm looking into the rabbitmq server api, and I don't see any cancellation mechanisms for a sent message so not sure how that would play out.

michaelklishin:
You cannot "unsend" a message in general but the use of context was discussed in the context of a general operation error handling approach, not something specific to publishes.

marwan-at-work:
@michaelklishin thanks for the timely replies! Any sort of context support is probably of good use. I'm happy to submit a PR. Haven't really dug into the source code, so I'd probably need some pointers. 

Also, here are some suggestions, just thinking out loud on github: 

1. maybe if we had a new function, something like: `ch.PublishWithContext(ctx, ex, routingKey...etc.)`
it wouldn't break the API. And likewise `ch.ConsumeWithContext`. Just a thought. either branching to `v2` or this are valid options imho.

2. I imagine implementing cancellation as the following: the client would declare a temporary Queue. If the context ever gets cancelled, it would send a cancel message on that queue. On the consumer side, there's a goroutine listening on that temporary Queue, if it receives the cancellation method it would update its state to cancelled. I think we can leave this up to the user to implement, but might also be interesting if the client took care of all of that boilerplate for the user.

michaelklishin:
@marwan-at-work I appreciate your willingness to help but this is a longer-term item at the moment. Which is not to say nobody should work on it but the maintainers won't get to it for a while. I think @streadway has a rough idea of how that might work. In case you are willing to invest some time into a prototype that may be thrown away, we would be very grateful and looking forward to see what comes out of it!

marwan-at-work:
@michaelklishin sounds good! I'll play with some ideas in my free time.

nvx:
[...]

PR #389, opened 2019-02-22
Implementation of Context based method for Channel

AlexisMontagne:
Related to #266 

Implement a second set of RPC method on the `Channel` having a context.Context as first argument


streadway:
This is great. To bikeshed the names a bit, I think the existing names like `QosContext` etc make sense because they follow the standard library adoption of context aware IO like `Dial` and `DialContext`.

With the async nature of client or server channel closure, cancelling the protocol is prone to subtle bugs. Please add some test coverage for the context cancellation cases.

AlexisMontagne:
@michaelklishin @streadway thanks for the feedback. I wanted to make sure the implementation would be ok with you.

I will add some tests and documentation for the new public methods.

To avoid out-of-sequence received messages, i cancel the channel in case of call cancelling. Do you thing to do?

michaelklishin:
OK, if this follows an existing naming convention then we should use those names.

@AlexisMontagne I'm not sure I understand the last question of yours, sorry.

AlexisMontagne:
@michaelklishin sorry, i submitted my comment by mistake. 

I meant: In order to deal with canceled RPC calls. Should we close the channel? or keep it open? cf. https://github.com/streadway/amqp/pull/389/files#diff-861e77702f49a71d32be4b3c7600818fR185

streadway:
> I meant: In order to deal with canceled RPC calls. Should we close the channel? or keep it open?

My first reaction is to keep the channel open. AMQP Channel lifecycle is complicated and shares client and server resources. If the context methods only relate to RPC round trips instead of server resources, it'd would be easier to document.

streadway:
When looking at error values, can you suggest a way of promoting a client side context.Context.Err() into an *amqp.Error?

AlexisMontagne:
@streadway you can find a standardized *amqp.Error  here: https://github.com/streadway/amqp/pull/389/commits/95dafc3f19f4cdbda1fe2ba4b756dc00808258dd

>> I meant: In order to deal with canceled RPC calls. Should we close the channel? or keep it open?
>
> My first reaction is to keep the channel open. AMQP Channel lifecycle is complicated and shares 
> client and server resources. If the context methods only relate to RPC round trips instead of server resources, it'd would be easier to document.

If we keep the channel open, i'm scared the message can cause issues

Let's say the following case:
T1: Channel X sends frame for RPC call Y 
T2: The context for the RPC call Y is canceled
T3: Channel X  sends frame for RPC call Z
T4: The dispatcher forward to the `rpc` channel the response for the RPC call Y

Is there any way to correlate request and response frame (sequence ID?)? In order to discard frame received for canceled calls

michaelklishin:
@AlexisMontagne other clients address this by not allowing channel sharing between threads/routines/etc. There are no sequence identifiers in the protocol, only response order guarantees for synchronous methods.

Note that sharing a channel for concurrent publishing, for example, is inherently unsafe because a published messages is 2 or 3 frames on the wire and you can end up with incorrect frame interleaving.

While closing the channel might be a decent option since there is no way to undo a send frame, the above makes me wonder if

 * Context support in this case would really achieve the goals it might in other libraries
 * Concurrent access to a channel should be advertised as supported (it may be in some cases for consumers but as with any mutable state, eliminating sharing avoids a lot of hard to dig out issues)

Keeping a channel open and documenting known limitations would be closer to what virtually every other RabbitMQ client does.

WDYT?

streadway:
> Note that sharing a channel for concurrent publishing, for example, is inherently unsafe because a published messages is 2 or 3 frames on the wire and you can end up with incorrect frame interleaving.

This library is concurrent safe for all messages, including publishing, by holding a channel specific mutex until all publish frames are sent before sending others.

streadway:
> * Context support in this case would really achieve the goals it might in other libraries

The primary use case for the context package is to represent a distributed, cancelable transaction.

It sounds like this is a discussion of "does a cancelled context propagate to the server or not?"

If we were to close a channel on cancellation during any RPC, would we also close the connection should we not be able to open a channel (RPC on channel id 0)?

Let's start with discarding the in-flight reply after cancellation to keep the channel open, and let the application determine how to propagate a cancellation to the server based on what the application considers a transaction.

michaelklishin:
Then it's up to the client to filter out the responses it cannot/no longer needs to handle. Java client and Bunny do this by yes, ignoring the responses they can detect as stale/mismatching the pending continuation (pending operation).

The only scenario where one cannot open a channel is if you run into the `channel_max` limit (`2047` in modern RabbitMQ versions by default) and that is pretty rare and usually indicates a channel leak. Closing a connection in that case would still be pretty disruptive and unexpected :/

AlexisMontagne:
@streadway i updated the pull request according to your last message. And added a test to ensure the messages received after the call is canceled are dropped.

connection.go
```diff
@@ -488,7 +488,10 @@ func (c *Connection) dispatch0(f frame) {
 				c <- Blocking{Active: false}
 			}
 		default:
-			c.rpc <- m
+			select {
+			case c.rpc <- m:
+			default:
```
streadway: Why does this now drop if there is no longer a receiver on the `rpc` buffered channel?

channel.go
```diff
@@ -340,7 +347,10 @@ func (ch *Channel) dispatch(msg message) {
 		// deliveries are in flight and a no-wait cancel has happened
 
 	default:
-		ch.rpc <- msg
+		select {
+		case ch.rpc <- msg:
+		default:
```
streadway: Same as on the connection, this seems like it could drop replies for unintended reasons.

channel.go
```diff
@@ -171,19 +171,26 @@ func (ch *Channel) open(ctx context.Context) error {
 func (ch *Channel) call(ctx context.Context, req message, res ...message) error {
 	select {
 	case <-ctx.Done():
-		return ctx.Err()
+		return ErrCanceled
 	default:
 	}
 
+	if req.wait() {
```
streadway: The reply from a cancelled request may be in-flight and not reside on the `rpc` at this point and it will not drain.

Take the scenario all with different contexts:

```
client send(a)
client send(b)
client wait(a) - cancelled - returned
server reply(a)
client wait(b) - recv reply(a)
```

AlexisMontagne: i don't think this use case is possible. Every RPC call is surounded by a mutex.

Therefore that could end up as:
```
client send(a)
client wait(a) - cancelled - returned
server reply(a) - message dropped
client send(b)
client wait(b) 
server reply(b) 
recv reply(b) 
```

Or (the case i exposed here https://github.com/streadway/amqp/pull/389#issuecomment-467143264)

```
client send(a)
client wait(a) - cancelled - returned
client send(b)
client wait(b) 
server reply(a) - message dropped
recv reply(a)  
```
And it causes issues.

But i can not figure out any way to handle this case with the current protocol. IMHO the safest and most deterministic way of dealing with this situation is to close the channel in case of cancelation.

AlexisMontagne:
@streadway what do you think of https://github.com/streadway/amqp/pull/389#discussion_r261221451?

channel.go
streadway: I still think the most obvious behavior is not to close the channel when a context is `Done()`. I believe context done is closer to a late `no-wait` rather than a client error.

This is a matter of sequencing the consumption of the replies in the same order as the requests were written. It looks like communication between the writer goroutine (application) and reader goroutine (dispatch) is how the sequencing would be implemented.

AlexisMontagne:
@streadway i think i got what you mean. Tell me what you think of that last commit.

Are we really sure the server will always reply to the client? Will it never give up on replying to any message? 

All my reasoning was based on the fact the server could be flaky and skip replies.

streadway:
AMQP requires TCP which offers transport reliability. The heartbeater will close the connection on read timeout, which will return an error should we experience any message loss.

I've found RabbitMQ very reliable and not drop or show flakey behavior, so we can count on every protocol request receiving a reply.

A counter is one approach, but that assumes that contexts will be cancelled in order which I don't think we would be able to depend on. Consider the message exchange:

```
req(a)
req(b)
req(c)
req(d)
cancel(a) 0->1
cancel(c) 1->2
reply(a) - dropped 2->1
reply(b) - dropped 1->0
reply(c) - dispatch(b)
reply(d) - dispatch(d)
```

AlexisMontagne:
@streadway not quite sure to get your example. One of the pre requisite of my implementation is to not have concurrent RPC calls and in fact there is mutex surounding the calls emitted by a specific `Channel`

AlexisMontagne:
@streadway for some reason i was thinking a was mutex surrounding every `call()` calls 🤔 . 

I updated my PR to handle the use case you described (and added a test to ensure it).

AlexisMontagne:
@streadway it should be good now. I wrapped the "sending" part of the `call` method in a mutex.

It also makes the frame sending thread safe. Previously https://github.com/streadway/amqp/blob/master/channel.go#L219 was not limited by any `Channel` wide mutex.

@michaelklishin The added tests should now pass with the `-race` options

michaelklishin:
@AlexisMontagne thank you for fixing the problematic test.

channel.go
```diff
@@ -642,7 +693,52 @@ greater as described by benchmarks on RabbitMQ.
...
+	return ch.QosContext(
+		context.Background(),
+		prefetchCount,
+		prefetchSize,
+		global,
+	)
+}
+
+/*
+QosContext controls how many messages or how many bytes the server will try to
```
michaelklishin: RabbitMQ does not support the byte limit. Let's leave it out.

channel.go
```diff
@@ -642,7 +693,52 @@ greater as described by benchmarks on RabbitMQ.
...
+started with noAck.
+
+When global is true, these Qos settings apply to all existing and future
+consumers on all channels on the same connection.  When false, the Channel.Qos
+settings will apply to all existing and future consumers on this channel.
+
+Please see the RabbitMQ Consumer Prefetch documentation for an explanation of
+how the global flag is implemented in RabbitMQ, as it differs from the
+AMQP 0.9.1 specification in that global Qos settings are limited in scope to
+channels, not connections (https://www.rabbitmq.com/consumer-prefetch.html).
```
michaelklishin: Please link to https://www.rabbitmq.com/confirms.html, which is the main guide on acknowledgements.

channel.go
```diff
@@ -642,7 +693,52 @@ greater as described by benchmarks on RabbitMQ.
...
+To get round-robin behavior between consumers consuming from the same queue on
+different connections, set the prefetch count to 1, and the next available
+message on the server will be delivered to the next available consumer.
+
+If your consumer work time is reasonably consistent and not much greater
+than two times your network round trip time, you will see significant
+throughput improvements starting with a prefetch count of 2 or slightly
+greater as described by benchmarks on RabbitMQ.
+
+http://www.rabbitmq.com/blog/2012/04/25/rabbitmq-performance-measurements-part-2/
```
michaelklishin: There's another blog post dedicated to this topic: https://www.rabbitmq.com/blog/2014/04/14/finding-bottlenecks-with-rabbitmq-3-3/. I'd suggest linking to it.

channel.go
```diff
@@ -746,6 +865,71 @@ declared with these parameters, and the channel will be closed.
...
+  Publish("", "alerts", false, false, Publishing{Body: []byte("...")})
+
+  Delivery       Exchange  Key       Queue
+  -----------------------------------------------
+  key: alerts -> ""     -> alerts -> alerts
+
+The queue name may be empty, in which case the server will generate a unique name
+which will be returned in the Name field of Queue struct.
+
+Durable and Non-Auto-Deleted queues will survive server restarts and remain
```
michaelklishin: A durable auto-delete queues that never had any consumers and doesn't have TTL applied to it will also survive a hosting node restart. I'd suggest linking to http://www.rabbitmq.com/queues.html instead of explaining it all here.

channel.go
```diff
@@ -746,6 +865,71 @@ declared with these parameters, and the channel will be closed.
...
+when there are no remaining consumers or bindings.  Persistent publishings will
```
michaelklishin: "Persistent messages".

channel.go
```diff
@@ -746,6 +865,71 @@ declared with these parameters, and the channel will be closed.
...
+be restored in this queue on server restart.  These queues are only able to be
+bound to durable exchanges.
+
+Non-Durable and Auto-Deleted queues will not be redeclared on server restart
```
michaelklishin: Again, auto-deletion is orthogonal to durability.

channel.go
```diff
@@ -746,6 +865,71 @@ declared with these parameters, and the channel will be closed.
...
+Durable and Auto-Deleted queues will be restored on server restart, but without
+active consumers will not survive and be removed.  This Lifetime is unlikely
+to be useful.
+
+Exclusive queues are only accessible by the connection that declares them and
+will be deleted when the connection closes.  Channels on other connections
+will receive an error when attempting  to declare, bind, consume, purge or
+delete a queue with the same name.
+
+When noWait is true, the queue will assume to be declared on the server.  A
```
michaelklishin: `noWait` simply means that there won't be a confirmation (a `queue.declare-ok` frame) sent back.

channel.go
```diff
@@ -786,6 +970,27 @@ can be used to test for the existence of a queue.
...
+		args,
+	)
+}
+
+/*
+
+QueueDeclarePassiveContext is functionally and parametrically equivalent to
+QueueDeclareContext, except that it sets the "passive" attribute to true.
+A passive queue is assumed by RabbitMQ to already exist, and attempting to
+connect to a non-existent queue will cause RabbitMQ to throw an exception.
```
michaelklishin: Clients connect to nodes, not queues. "Attempting to passively declare a non-existent queue will result in a channel-level protocol exception".

channel.go
```diff
@@ -832,13 +1037,32 @@ channel will be closed.
...
+/*
+QueueInspectContext passively declares a queue by name to inspect the current
+message count and consumer count.
+
+Use this method to check how many messages ready for delivery reside in the queue,
+how many consumers are receiving deliveries, and whether a queue by this
+name already exists.
+
+If the queue by this name exists, use Channel.QueueDeclare check if it is
+declared with specific parameters.
```
michaelklishin: With matching parameters. There's a [separate doc section](https://www.rabbitmq.com/queues.html#property-equivalence) that covers this topic.

channel.go
```diff
@@ -894,11 +1118,67 @@ closed with an error.
...
+		key,
+		exchange,
+		noWait,
+		args,
+	)
+}
+
+/*
+QueueBindContext binds an exchange to a queue so that publishings to the
+exchange will be routed to the queue when the publishing routing key matches the
```
michaelklishin: Some exchanges do not use routing keys or the idea of "matching" is quite a stretch. Should we link to https://www.rabbitmq.com/getstarted.html instead?

AlexisMontagne:
@michaelklishin i did not write any fo this documentation. I used the one already in the code for the context-free methods

michaelklishin:
@AlexisMontagne OK, thanks. I will revisit that together with the new functions then.

michaelklishin:
@streadway any opinion on this final version?

channel.go
```diff
@@ -161,25 +166,45 @@ func (ch *Channel) send(msg message) (err error) {
...
+	if req.wait() {
+		ch.responses = append(ch.responses, &status)
+	}
+
+	ch.responsesM.Unlock()
+
 	if req.wait() {
 		select {
+		case <-ctx.Done():
+			atomic.StoreUint32(&status, 1)
```
streadway: There is a race conditions between receiving `ctx.Done()`, storing `status = 1` and loading `status` in `dispatch()`.

Instead, let's collect the in-flight contexts, so we will coordinate by communicating over `Context.Done()` rather than sharing the `status` state.

I think this would mean a responses slice of `[]atomic.Value` and storing and loading `context.Context` types.

channel.go
```diff
@@ -330,7 +355,21 @@ func (ch *Channel) dispatch(msg message) {
 		// deliveries are in flight and a no-wait cancel has happened
 
 	default:
-		ch.rpc <- msg
+		var status *uint32
+
+		ch.responsesM.Lock()
+
+		if len(ch.responses) == 0 {
```
streadway: This *should* never happen. I think we ought to panic to reveal any bugs.

channel.go
```diff
@@ -330,7 +355,21 @@ func (ch *Channel) dispatch(msg message) {
...
+			ch.responsesM.Unlock()
+			return
+		}
+
+		status, ch.responses = ch.responses[0], ch.responses[1:]
+		ch.responsesM.Unlock()
+
+		if atomic.LoadUint32(status) == 0 {
```
streadway: If we were using channels instead of state, this would be:

```
switch {
  case <-status.Done():
  default:
    ch.rpc <- msg
}
```

streadway:
> I admire the amount of effort you've put into the docs and perhaps 50 to 65% of the docs can be retained. Unfortunately the docs are not always correct and reinvent RabbitMQ doc guides where it's not really necessary. I'd recommend linking to the relevant RabbitMQ guides more. If something is missing in the guides (e.g. a guide on bindings besides the tutorials) I'd be happy to work on that.
This will benefit all RabbitMQ users regardless of the client.

@michaelklishin This would be great! The method docs were written in 2012 and tried to strike the balance between sharing enough context about AMQP principles, RabbitMQ extensions/constraints and Client library choices. I'd very much welcome documentation review and improvements from the RabbitMQ team!

channel.go
streadway: Alternatively, the responses slice could contain a buffered chan of messages and that would be used instead of `rpc`.

michaelklishin:
@streadway and I discussed this on a Hangout. Besides splitting this PR into a few (@AlexisMontagne thank you for your truly exemplary patience with us!), we'd like to revisit the internal concurrency model used by this client to both simplify `Context` adoption and make it closer to other clients. It has a chance of improving CPU resource utilisation on some workloads, too ;)

@streadway is to file a new issue about that but basically operation dispatch should be synchronised per channel and not over the entire connection. Channels are assumed to be "independent entities" in the protocol. Publishing on a shared [protocol] channel is not supported so other clients guarantee total ordering on a channel, not across channels. A single channel in the server never dispatches its operations concurrently either. This would make `Context` support more straightforward.

michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

Issue #390, opened 2019-03-05
Example: Connection not closed if channel exception is not connection exception


Raqbit:
The [Example](https://github.com/streadway/amqp/blob/1fcab46dce7d4148a84fc6eb15898f6865c7df6f/example_client_test.go#L88) uses `NotifyClose()` on the channel to detect when a connection/channel exception occurs, then reconnects to RabbitMQ & recreates the channel. By doing this it depends on the broadcasting of connection exceptions to all channel exception listeners.

Now, I might be missing something here, but if an _actual_ channel exception occurs:

1. We are creating a new connection even though the connection might be fine
2. We are not closing the old connection and are creating a new one

Why is the example using `Channel.NotifyClose()` this way, and is this best practice?

michaelklishin:
Channel exceptions generally do not trigger connection recovery. Simply opening a new channel in that case is enough. So this sounds like a mistake that should be corrected. Please submit a PR?

Raqbit:
Closed by #392

PR #392, opened 2019-03-06
Handle channel exceptions in example client

Raqbit:
Solves #390.

The client example is now able to handle channel exceptions and will recreate the channel once one occurs.

### DIFF
--- a/example_client_test.go
+++ b/example_client_test.go
@@ -27,7 +27,7 @@ func Example() {
 	for {
 		time.Sleep(time.Second * 3)
 		if err := queue.Push(message); err != nil {
-			fmt.Printf("Push failed: %s", err)
+			fmt.Printf("Push failed: %s\n", err)
 		} else {
 			fmt.Println("Push succeeded!")
 		}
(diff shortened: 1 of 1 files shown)


example_client_test.go
```diff
@@ -36,28 +36,32 @@ func Example() {
...
+
 	// When resending messages the server didn't confirm
 	resendDelay = 5 * time.Second
 )
 
 var (
 	errNotConnected  = errors.New("not connected to the queue")
-	errNotConfirmed  = errors.New("message not confirmed")
 	errAlreadyClosed = errors.New("already closed: not connected to the queue")
+	errShutdown      = errors.New("queue is shutting down")
```
michaelklishin: This is a good chance to correct this poorly worded message. Clients do not connect to queues, they [connect to nodes](https://www.rabbitmq.com/connections.html).

example_client_test.go
```diff
@@ -36,28 +36,32 @@ func Example() {
 
 // Queue represents a connection to a specific queue.
```
michaelklishin: I know it's not in the original scope but this badly needs renaming. Clients do not connect to queues. `ConsumerState` would be a better name here.

example_client_test.go
```diff
@@ -36,28 +36,32 @@ func Example() {
...
+	notifyConfirm   chan amqp.Confirmation
+	isReady         bool
 }
 
 const (
 	// When reconnecting to the server after connection failure
 	reconnectDelay = 5 * time.Second
 
+	// When setting up the channel after a channel exception
+	resetupDelay = 2 * time.Second
```
michaelklishin: "resetup" is not really a word. I think "reinitialization" or "reinit" would be clearer and more familiar here.

example_client_test.go
```diff
@@ -73,35 +77,86 @@ func New(name string, addr string) *Queue {
...
-		return false
+		return nil, err
+	}
+
+	queue.changeConnection(conn)
+	log.Println("Connected!")
+	return conn, nil
+}
+
+func (queue *Queue) handelResetup(conn *amqp.Connection) bool {
```
michaelklishin: Same as above, and `handel` should probably be `handle`.

example_client_test.go
```diff
@@ -73,35 +77,86 @@ func New(name string, addr string) *Queue {
...
+	for {
+		queue.isReady = false
+
+		err := queue.setup(conn)
```
michaelklishin: `ConsumerState.init` or `AppState.init` would fit quite well here.

example_client_test.go
Raqbit: What do you think of `Session`?

michaelklishin: `Session` is pretty good.

michaelklishin:
@Raqbit thank you for updating the names and being so patient.

PR #397, opened 2019-03-17
Internal proto package

streadway:
This introduces internal packages, starting with the protocol message generation/serialization and later for common testing, supportive data structures and possible alternative implementations.

Changes to the `amqp` package interface includes aliasing some constants, variables, types and errors to `github.com/streadway/amqp/internal/proto`. These should be backwards compatible exports.

Everyone is welcome to review, please include comments on areas that need focus or an application integration test.


michaelklishin:
@streadway I haven't finished reviewing yet but I like this so far. I had to correct what looked like a typo for this branch to build: type `frame` had to be `Frame` in the generated spec parser.

michaelklishin:
@streadway I think this is ready to go. Shall I merge?

PR #399, opened 2019-04-14
Add preallocation for content buffer

denperov:
This significantly reduces memory allocation for large messages.


michaelklishin:
Do you have any benchmarks or at least numbers to demonstrate the difference?

michaelklishin:
Thank you!

PR #404, opened 2019-05-15
print validate not support type

xieyanker:
if argument of validateField function is not supported  type, such as `uint32`,
 I got: `table field "x-message-ttl" value %!t(uint32=300000) not supported`,
 now I fix it and print actually type: ` table field "x-message-ttl" value uint32 not supported`.


michaelklishin:
Thank you!

Issue #403, opened 2019-05-14
Creating a pool of channels?


xamenyap:
Hi,

I am trying to find a way to create a pool of channels on one connection. As I could not find any examples, I tried to come up with my own implementation. Here it is:

```go
type Publisher struct {
	url           string
	pollDuration  time.Duration
	channelsCount int
	conn          *amqp.Connection
	sessions      []session
	errorCh       chan error
	amqpErrorCh   chan *amqp.Error
}

type session struct {
	channel *amqp.Channel
	confirm chan *amqp.Confirmation
}

func NewPublisher(
	url string,
	pollDuration time.Duration,
	channelsCount int,
) *Publisher {
	p := &Publisher{
		url:           url,
		pollDuration:  pollDuration,
		channelsCount: channelsCount,
		errorCh:       make(chan error),
		amqpErrorCh:   make(chan *amqp.Error),
	}

	go func() {
		for {
			select {
			case <-p.amqpErrorCh:
				p.reconnect()
			case <-p.errorCh:
				p.reconnect()
			}
		}
	}()

	go p.connect()

	return p
}

func (p *Publisher) connect() {
	conn, err := amqp.Dial(p.url)

	if err != nil {
		p.errorCh <- err
		return
	}

	conn.NotifyClose(p.amqpErrorCh)

	p.conn = conn
	p.sessions = make([]session, 0)

	for len(p.sessions) < p.channelsCount {
		ch, err := conn.Channel()
		if err != nil {
			p.errorCh <- err
			return
		}

		p.sessions = append(p.sessions, session{ch, make(chan *amqp.Confirmation)})
	}
}

func (p *Publisher) reconnect() {
	time.Sleep(p.pollDuration)
	p.connect()
}
```

The basic idea here is to create a number of sessions, which holds an __*amqp.Channel__ and and a __chan *amqp.Confirmation__ as I need to use confirm mode in my project.
A go routine is spawned in __NewPublisher()__ in order to listen to the errors, and re-create another connection and a batch of channels.
I really appreciate it if someone can give me some reviews on this implementation. I can contribute to the examples if this is an appropriate approach. Thanks!

hsyed:
I need something similar and this is one way of solving the problem.  I'm not that familiar with AMQP. @streadway should it not be possible to multiplex multiple publish confirms over a single channel ?.

hsyed:
| What are you trying to accomplish with the pool of channels?

Reliable publishes, this is the abstraction I created for this:

```
[...]

PR #424, opened 2019-11-05
Add an trivial example on how to establish a pool of publishers

xamenyap:
Follow up on #403 I made this PR to demonstrate how to create a pool of publishers on 1 connection. The example focuses on how the pool can be constructed and used, many details are omitted such as:
- Error handling
- Publish confirm handling
- Connection redialing
- etc.


michaelklishin:
Thank you for contributing. I'm afraid that exactly because so many practically important parts are missing, we should leave this specific example out. You are more than welcome to publish it in a separate repo of your own, of course.

PR #426, opened 2019-11-20
Channel buffer size is accessed through `cap` not `len`

puellanivis:
It’s really confusing that for channels, it is `make(chan T, capacity)`. So that `len(chan T)` gives how many items are in the buffer, while `cap(chan T)` gives the buffer size.

As a result, we almost certainly do not have anything queued onto the `ack` and `nack` channels already, which means these `len(…)` calls are almost certainly always zero.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -553,7 +553,7 @@ ordered Ack and Nack DeliveryTag to the respective channels.
 For strict ordering, use NotifyPublish instead.
 */
 func (ch *Channel) NotifyConfirm(ack, nack chan uint64) (chan uint64, chan uint64) {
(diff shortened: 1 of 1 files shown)


puellanivis:
```
	integration_test.go:1803: dial integration server: dial tcp 127.0.0.1:5672: connect: connection refused
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

Hm… the `t.Errorf("dial integration server: %s", err)` should be `t.Fatal("dial integration server:", err)` so that the test immediately fails rather than panics.

streadway:
Thanks for the contribution! To save some time, I cherry picked your commit on top of the fixed build to get this merged faster. Please re-fork if you'd like to pickup your change again.

Closed by #430

PR #440, opened 2020-03-09
fix typo

mind1949:


example_client_test.go
```diff
@@ -26,7 +26,7 @@ func Example() {
 	message := []byte("message")
 	// Attempt to push a message every 2 seconds
 	for {
-		time.Sleep(time.Second * 3)
+		time.Sleep(time.Second * 2)
```
michaelklishin, line 29: Hmmm, I guess this is a harmless unrelated change in practice.

PR #442, opened 2020-04-08
feat: implement options pattern for to simplify Dial API

troian:
Implements #442


troian:
all should be fixed now

PR #456, opened 2020-06-22
c.reader() not always exit on c.Close()

makasim:
On c.Close() c.heartbeater() might exit just before c.deadlines <- conn is called. Which results in this goroutine being stuck forever.


michaelklishin:
Thank you!

PR #463, opened 2020-07-15
Increase publish performance of small message ~500%

carlhoerberg:
By writing all 3 frames required for a publish in one go, and only
locking and flushing the output buffer once we increase the performance
about 3 times.

Messages that spans over multiple body frames are still written one at a
time, when messages are that large (>128KB) the locking and flushing is
not the bottleneck, and it allows us to not allocate a dynamic array for
each publish and allows us to use a fixed size array instead.

By disabling TCP no delay (enable Nagle's algorithm) many small messages
can be sent in a single TCP packet, we increase the publish rate about
100% when messages are small.

This will not increase latency in any normal circumstances, but
theoretically could if for instance one channel is publishing a message,
and another channel is declaring a queue and then waiting for the CreateOK
response, then due to a delayed ack from the server a 40ms delay could
be added to the wait of the Queue CreateOK.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -229,21 +229,18 @@ func (ch *Channel) sendOpen(msg message) (err error) {
 		} else {
 			size = len(body)
 		}
+		var frames [3]frame
 
-		if err = ch.connection.send(&methodFrame{
+		frames[0] = &methodFrame{
 			ChannelId: ch.id,
 			Method:    content,
-		}); err != nil {
-			return
 		}
 
-		if err = ch.connection.send(&headerFrame{
+		frames[1] = &headerFrame{
 			ChannelId:  ch.id,
 			ClassId:    class,
 			Size:       uint64(len(body)),
 			Properties: props,
-		}); err != nil {
-			return
 		}
 
 		// chunk body into size (max frame size - frame header size)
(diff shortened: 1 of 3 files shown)


michaelklishin:
RabbitMQ Java client [sets `TCP_NODELAY` to `true` by default](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/DefaultSocketConfigurator.java#L36).
RabbitMQ .NET client [sets `TCP_NODELAY` to `true` by default](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/2a709aa76f25e8b6754cef4e4f482547d6c1d870/projects/RabbitMQ.Client/client/api/ConnectionFactoryBase.cs#L54)
FWIW Netty [sets it to true by default](https://netty.io/5.0/api/io/netty/channel/socket/SocketChannelConfig.html#isTcpNoDelay()), and it's a key building block for a lot of network protocol clients our there.

So I am not convinced this client should adopt a different default. It can make it easy for the user to control the value. That would be something pretty unrelated to how framesets are written to the socket.

michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

Issue #476, opened 2020-09-29
Publish method does not return "confirms.Published" counter(deliveryTag)


kamal-github:
I have a listener registered through `NotifyPublish` and it is asynchronously getting confirmation and if ack=false then I republish the message or act upon it, but I get delivery tag from this listener channel and I do not know which payload this delivery tag is of.

While looks like we can return the https://github.com/streadway/amqp/blob/v1.0.0/confirms.go#L36 which looks like the delivery tag for the given channel, but we are ignoring it in Publish method https://github.com/streadway/amqp/blob/v1.0.0/channel.go#L1360.

chrisDeFouRire:
Hi @kamal-github 

I was wondering the same, until I remembered that confirmations are sent on the channel where I publish, only for what I publish on that channel... 

ie. confirmation notification with Tag=1 is for the first message you published on the channel, etc... It's easy to keep track of it (but I agree, Publish could have returned that counter).

kamal-github:
@chrisDeFouRire I am worried about maintaing the same counter which might have concurency issue. How about adding a method like Java client has. `getNextSequenceNumber()` ? https://www.rabbitmq.com/tutorials/tutorial-seven-java.html

chrisDeFouRire:
@kamal-github I like your solution, though I can only give it my vote 👍

michaelklishin:
This client uses a really conservative approach when it comes to even theoretically breaking public API changes. Exposing a sequence counter reader — is the way to go.

michaelklishin:
Please submit a PR that introduces a new function that returns the counter, or a doc update if it already exists. Breaking `Channel.Publish` modifications (such as the return value) are out of the question for this client.

chrisDeFouRire:
@michaelklishin I think that's exactly what PR #478 (by kamal-github) does

kamal-github:
@michaelklishin As @chrisDeFouRire mentioned that https://github.com/streadway/amqp/pull/478 I have already created the PR for this. Please review. :)

qiaohongbo:
@michaelklishin 

When merging it. https://github.com/streadway/amqp/pull/478

qiaohongbo:
> @michaelklishin As @chrisDeFouRire mentioned that #478 I have already created the PR for this. Please review. :)

When merging it

samyonr:
[...]

PR #478, opened 2020-10-09
Add GetNextPublishSeqNo() for Channel in confirm mode

kamal-github:
`GetNextPublishSeqNo` method on `Channel` will be very useful for handling of especailly NACKed messages to keep track of message going to be published so that those can be republished or acted upon. The `confirms.published` counter is already there and this method will just return the same counter.

Resolves https://github.com/streadway/amqp/issues/476


chrisDeFouRire:
👍

qiaohongbo:
@michaelklishin 

When merging it. https://github.com/streadway/amqp/pull/478

channel.go
```diff
@@ -1290,6 +1290,12 @@ func (ch *Channel) ExchangeUnbind(destination, key, source string, noWait bool,
 	)
 }
 
+// GetNextPublishSeqNo returns the sequence number of the next message to be
+// published, when in confirm mode.
+func (ch *Channel) GetNextPublishSeqNo() uint64 {
+	return ch.confirms.published
```
fho, line 1296:
```suggestion
func (ch *Channel) GetNextPublishSeqNo() uint64 {
	c.m.Lock()
	defer c.m.Unlock()
	return ch.confirms.published
```

doesn't the lock needs to be hold?

chrisDeFouRire: `ch.confirms.Lock()` and `defer ch.confirms.Unlock()` would work better yet, or add a function in confirm?

fho, line 1296: According to the description the function should return the sequence number (== deliveryTag?) of the next message that is published.
But if `GetNextPublishSeqNo` is called before the first `Publish()` call it returns 0, the deliveryTag of the first publishing is 1 though

kamal-github: Yes, it makes sense to add `Lock` and `Unlock` for confirm. though it is Read operation, currently we are using `sync.Mutex` not `RWMutex` https://github.com/streadway/amqp/blob/master/confirms.go#L7 but I change it to `RWMutex` which is not going to impact any current functionality.

```
// Lock acquire the lock on confirms.
func (c *confirms) Lock() {
	c.m.Lock()
}

// Unlock release the locks on confirms.
func (c *confirms) Unlock() {
	c.m.Unlock()
}

Comment (kamal-github) [reply channel.go:1296]: You are right @fho I am gonna make change as 
```
func (ch *Channel) GetNextPublishSeqNo() uint64 {
	ch.confirms.Lock()
	defer ch.confirms.Unlock()
	
	return ch.confirms.published + 1
}

fho: I would keep the mutex. It should be faster when the lock is only hold very short timespans, which is the case here.
It might be much slower with an rwmutex  (https://github.com/golang/go/issues/38813)

fho: confirms.expected can't be used instead? I only had a quick look and was not sure :-)

kamal-github: `published` incremented with every publish on the given channel. `expecting` increments, only when confirms are received as a part of resequencing and RabbitMQ broker does not guarantee that how soon a message will be confirmed. so `published` and `expecting` may be different at one moment and expecting may leg behind. Please correct me If I am wrong. :)

fho: Thank you for checking.  I agree, `expecting` is only incremented when a confirmation was received not for every publish call.

michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

samyonr:
For everyone who reaches this thread now, note that the exact same commit was released in `rabbitmq/amqp091-go`: https://github.com/rabbitmq/amqp091-go/releases/tag/v1.2.0

PR #496, opened 2021-02-01
Channel shutdown process improvements: shutdown error added as an own…

askretov:
**Problem:**
Ambiguous behavior on Channel shutdown. If you consume using a Channel you cannot get the shutdown cause even if you use NotifyClose chan listener. Delivery chan could be selected (closed) earlier than close listener chan.

**Scenario:**
Let's assume you have a connection and few consumer channels using it.
The requirement is to restart the consumer job on any unexpected error... **on unexpected errors/shutdowns only**.
When your connection closes from outside (unexpected error or gracefully closes from parallel goroutine) you have to handle a shutdown cause distinctly with no doubt.
Look at this sample pseudo-code first:

```
delivCh, err := ch.Consume(...someConf...)
// some error processing
closeCh := ch.NotifyClose(make(chan *amqp.Error))
for {
	select {
	case msg, ok := <-delivCh:
		if !ok {
			// As we understand, delivCh could be selected earlier than closeCh and sometimes it works
                        // right in this way... but it's ok, it's normal behavior of select.
			// Deliveries chan is closed, but what's a cause? It could be a graceful shutdown then 
                        // we need no restart.
			// That's why such behaviour is ambiguous now.
		}
		// Process the message
	case err, ok := <-closeCh:
		if !ok {
			log.Println("channel closed gracefully")
			// Return nil
		} else if err != nil {
			log.Println("channel going to shutdown with an error")
			// Return an error, restart the job
		}
	}
}
```

In this scenario you cannot guarantee that closeCh will be selected early than delivCh.

**PR proposal:**
To store a shutdown cause error into a Channel's field shutdownErr upon shutdown start and provide a getter method to get it any time at any place during shutdown process.
New instance of Channel has default value for shutdownErr with "unexpected shutdown" Reason to provide a small guard for the future. It will not affect current behaviour because it's being set upon shutdown start, but if someday somebody will try to close the chans bypassing shutdown() func it will return a distinct error.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -72,6 +72,9 @@ type Channel struct {
 	message messageWithContent
 	header  *headerFrame
 	body    []byte
+
+	// Shutdown cause error
+	shutdownErr *Error
 }
 
 // Constructs a new channel with the given framing rules
@@ -84,6 +87,9 @@ func newChannel(c *Connection, id uint16) *Channel {
 		confirms:   newConfirms(),
 		recv:       (*Channel).recvMethod,
 		errors:     make(chan *Error, 1),
+		shutdownErr: &Error{
+			Reason: "unexpected shutdown",
+		},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Hey folks,

involved.

reasonably quickly).

as practically possible soon.

PR #508, opened 2021-06-16
fix #254 Channel.Publish blocks indefinitely

### LINKED ISSUE #254: Channel.Publish blocks indefinitely
Channel.Publish sometimes blocks indefinitely. The following code is a test case to reproduce this. It reproduces most of time I run it. Adding a sleep between inside the for loop does not help.

```
package main

import (
	"log"

	"github.com/streadway/amqp"
)

func main() {

	const queueName = "q1"

	if conn, stdErr := amqp.Dial("amqp://guest:guest@localhost:5672/"); stdErr != nil {
		log.Fatalln(stdErr)
	} else if channel, stdErr := conn.Channel(); stdErr != nil {
		log.Fatalln(stdErr)
	} else if stdErr := channel.Confirm(false); stdErr != nil {
		log.Fatalln(stdErr)
	} else if _, stdErr := channel.QueueDeclare(
		queueName, //queue
		true,      //durable
		false,     //autodelete
		false,     //exclusive
		false,     //nowait
		nil,       //arguments
	); stdErr != nil {
		log.Fatalln(stdErr)
	} else {
		confirmationChan := channel.NotifyPublish(make(chan amqp.Confirmation))

		for i := 0; i < 10000; i++ {
			log.Println("publish ", i)

			if stdErr := channel.Publish(
				"",        // exchange
				queueName, // routing key
				false,     // mandatory
				false,     // immediate
				amqp.Publishing{
					Body: []byte{},
				}); stdErr != nil {
				log.Fatalln(stdErr)
			} else {
				log.Println("pre confirmation check")
				confirmation := <-confirmationChan
				log.Println("post confirmation check")
				if !confirmation.Ack {
					log.Fatalln("confirmation not ack")
				}
			}
			// time.Sleep(time.Millisecond * 1000)
		}
	}
}
```

sljeff:
I encountered a bug similar to this comment: https://github.com/streadway/amqp/issues/254#issuecomment-316401642

However, in a weak network environment, I get a deadlock after N+1 messages have been sent where N is the size of the buffered channel.

---

How to Reproduce:
- weak network enviroment (I was using `Network Link Conditioner` with `Very Bad Network` config)
- An external public network rabbitmq
- 10 goroutines sending messages to rabbitmq (I was using this library
https://github.com/shanbay/gobay/blob/v0.15.0/extensions/busext/amqp.go#L127)

Then it will deadlock.

---

The sequence of events is:
1. Receive an old message's ack -> `One()` -> `resequence()` -> `confirm()` send ack to channel
2. Receive another old message's ack -> `One() get lock` ->  `resequence()` -> `confirm()` send ack to channel -> block (channel buffer == 1) (and lock `confirms.m` is not released)
3. I am sending a message using `Publish` -> Failed to get lock

This time it will keep blocking; simply because the buffer is 1 but two ack are received at the same time (although I sent messages serially, we can indeed reproduce this extreme case).

---

I don't think publish needs to use the same lock as the other methods, it's just a self-incrementing. Even if the program is in `resequence()`, there may not be a problem with `Publish` at this point.

### DIFF
--- a/confirms.go
+++ b/confirms.go
@@ -1,13 +1,16 @@
 package amqp
 
-import "sync"
+import (
+	"sync"
+	"sync/atomic"
+)
 
 // confirms resequences and notifies one or multiple publisher confirmation listeners
 type confirms struct {
 	m         sync.Mutex
 	listeners []chan Confirmation
 	sequencer map[uint64]Confirmation
-	published uint64
+	published *uint64
 	expecting uint64
 }
 
@@ -15,7 +18,7 @@ type confirms struct {
 func newConfirms() *confirms {
 	return &confirms{
 		sequencer: map[uint64]Confirmation{},
-		published: 0,
+		published: new(uint64),
 		expecting: 1,
 	}
 }
@@ -29,11 +32,7 @@ func (c *confirms) Listen(l chan Confirmation) {
 
 // publish increments the publishing counter
 func (c *confirms) Publish() uint64 {
-	c.m.Lock()
-	defer c.m.Unlock()
-
-	c.published++
-	return c.published
+	return atomic.AddUint64(c.published, 1)
 }
 
 // confirm confirms one publishing, increments the expecting delivery tag, and
@@ -48,7 +47,7 @@ func (c *confirms) confirm(confirmation Confirmation) {
 
 // resequence confirms any out of order delivered confirmations
 func (c *confirms) resequence() {
-	for c.expecting <= c.published {
+	for c.expecting <= atomic.LoadUint64(c.published) {
 		sequenced, found := c.sequencer[c.expecting]
 		if !found {
 			return


nikitasadok:
I have had a very similar issue lately. There isn't anything special about the reproducing, there would just be a random deadlock which would prevent all the other `Publish` to this channel. It would lock up for at least an hour (didn't check longer time periods, but I am pretty sure that the deadlock would still be there). 
However, there was still one weird problem. At about the time when the deadlock appeared, I could see the following log in RabbitMQ Container:  [error] <...> closing AMQP connection <...>, so I am assuming that some connection problems could have caused this mixup in acks and lead to deadlock.
I completely agree with this pull request as I think that semantically Publish has to have another lock since it is just an increment which can be done more efficiently as well as more clearly with atomic operation.

sljeff:
@michaelklishin @streadway

hj24:
same question here. @michaelklishin @streadway

PR #515, opened 2021-09-09
Fix AMQPLAIN authentication mechanism

### LINKED ISSUE #514: AMQPLAIN authentication does not work
Hi,

I've encountered an issue when trying to use AMQPLAIN authentication. It seems that the current implementation does not work out of the box.

I looked for an official reference for the AMQPLAIN authentication but could not find any, but other implementations have different wire formats.

To reproduce, run [this gist](https://gist.github.com/hodbn/5b8fe5242a5e46e1c635108a21ffe869) against a `rabbitmq` container:
```
% docker run -d -p 5672:5672 rabbitmq
% go run main.go guest guest amqp://127.0.0.1:5672/
2021/09/09 10:51:38 can't dial: Exception (403) Reason: "username or password not allowed"
```

I'll add a PR later.

hodbn:
Fixes #514.

The current AMQPLAIN implementation does not work with `rabbitmq` out-of-the-box.

I could not find an official reference for that authentication mechanism. Looking at other implementations ([rust](https://github.com/CleverCloud/amq-protocol/blob/master/protocol/src/auth.rs#L45), [python](https://github.com/celery/py-amqp/blob/7300741f9fc202083e87abd10e1cb38c28efad92/amqp/sasl.py#L48)) it seems that AMQPLAIN uses an AMQP table (`LOGIN` and `PASSWORD` fields), omitting the table length.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -6,6 +6,7 @@
 package amqp
 
 import (
+	"bytes"
 	"fmt"
 )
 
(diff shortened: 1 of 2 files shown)


hodbn:
@michaelklishin Note, this was merged (and seems to work) on rabbitmq/amqp091-go#16.

michaelklishin:
@hodbn Team RabbitMQ no longer pays any attention to this fork. This is abandonware, https://github.com/rabbitmq/amqp091-go/ is where all new development takes place.

PR #535, opened 2023-04-21
Add deprecation warning to go.mod

nemith:
Following the instructions here https://go.dev/ref/mod#go-mod-file-module-deprecation
this will deprecate this module and point people to the supported version.

This will show up in linters and with `go mod` command as well as in the
pkg.go.dev website.


nemith:
I am submitting this as I didn't notice the deperecation in the readme until almost a year later just trying to find some documentation.  Adding this will allow linters, `go mod`, and pkg.go.dev to make it very clear that this isn't supported anymore.

PR #541, opened 2023-09-07
Error because the url contains special characters

lfxnxf:
Error because the url contains special characters;
For example, amqp://guest:guest#123@localhost:5672/
Call amqp.dial (url), return error :parse "amqp://guest:guest": invalid port ":guest" after host
This error is caused by url.Parse.

Add a DailURI method that uses URI connections to pass in connection information without error if special characters are included

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -224,6 +224,10 @@ func DialUri(uri *URI) (*Connection, error) {
 	})
 }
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Team RabbitMQ no longer pays any attention to this fork. Consider switching to, and submitting your PRs again, https://github.com/rabbitmq/amqp091-go/.

